PR #33, opened 2020-06-30
Adding location attribute

Samrthi:
Created location attribute for the class. Added column that shows the location that can be hidden. Backwards compatible.


src/app/person.ts
```diff
@@ -16,6 +16,7 @@ export class Person {
     constructor(
```
amdw, line 16: Please make sure the copyright date range in the header of each file you have touched extends to 2020, like this:

https://github.com/google/peoplemath/blob/37da3686c7e2fdac2b77fb130233b104ca610562/src/app/objective.ts#L1

I bet there's an editor plugin or something out there that can do this for you automatically.

Samrthi:
Hi Andrew,

Thanks for reviewing the pull request.
I totally missed the fact that I should be running the Angular tests, I only ran the Go tests, thanks for pointing that out! I was confused why the tests didn’t cover much of the front end 🤦🏼‍♀️🤦🏼‍♀️ I’ve got much to learn when it comes to testing so I’m sorry for making silly mistakes until I learn better 😬

I agree with the change to only show the column when at least one person has a location set, I will change that.
I didn’t touch the package-lock file so I’m not sure why that changed at all, that’s definitely a change that I shouldn’t have pushed.

Cheers,
Sam

amdw:
No worries Sam! The test coverage in the front-end isn't amazing right now, so the tests might well have passed anyway.

Mostly I add specific tests for the front-end only when there is an actual bug I'm trying to fix (which is why I suggested one in the context of the `Person` constructor).

It would be great to have more coverage of the features, but I've always been concerned that front-end tests might be quite fragile (i.e. tend to fail often when we're deliberately making small UI changes, more often than telling us when we have accidentally broken something).

I'm no Angular expert, though, and I'm very much open to the possibility that there is something better we can do here.

Regardless, it's a good idea to have `ng test` running in the background while you develop. It will re-run the test suite every time you make a code change. You can run it in parallel with `ng serve`, though it's better to let one of them fully start up before you start the other one, as I believe some of the startup operations take some sort of lock.

I'm definitely going to confirm that `ng test` and the Go backend tests are 100% green before I accept any PR. Ideally we'd have automation for that in the GitHub workflow, but we don't at the moment.

I agree, it's probably best to leave `package-lock.json` out of the PR unless we're doing a deliberate dependency change.

Another (separate) exercise that's probably worth you doing is to try to run an `ng update` to upgrade to the latest version of Angular ([docs](https://angular.io/cli/update)). That will definitely cause changes to `package-lock.json`.

backend/main.go
```diff
@@ -103,6 +103,7 @@ type Assignment struct {
 type Person struct {
 	ID           string  `json:"id"`
 	DisplayName  string  `json:"displayName"`
+	Location     string  `json:"location"`
```
amdw, line 106: One thing I forgot before: can you add a test to `main_test.go` to assert that you can set and retrieve this field in the JSON as you expect?

You needn't write a whole new test case - you can just add an assertion to `TestPostPeriod`, similar to what I did for objective notes in commit https://github.com/google/peoplemath/commit/99bf1520cb64504a669e76e2736ce37d2512e6d1.

amdw:
Unfortunately it looks like the recent commits with my changes applied have a `Co-authored-by:` value in the commit message that uses my personal email address. This is causing the CLA check to fail (since I haven't signed the CLA in my personal capacity).

Please could you either make sure the `Co-authored-by:` in your commit messages is set to my google.com email address, or omit it altogether? Thanks!

(The current values may be getting set because of some automated GitHub behaviour when you accept suggested changes. The primary email address associated with my GitHub account is my personal one. I don't see a way to request that an alternate email address be used when suggesting changes.)

You may need to [squash your work so far into a single commit](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History) with a new commit message. If you need to create a new pull request to do that, then that's fine.

src/app/people/people.component.ts
```diff
@@ -50,16 +51,24 @@ export class PeopleComponent implements OnInit {
...
   @ViewChild(MatSort) sort: MatSort;
-  
+
   constructor(public dialog: MatDialog) { }
 
   ngOnInit() {
   }
 
+  displayedColumns(): string[] {
+    for (var i = 0; i < this.people.length; i++) {
```
amdw, line 62: A few things here.

* Firstly, let's please avoid the repetition of elements between the "with location" and "without location" arrays.
* Secondly, instead of this for loop, it's more concise and readable to use the `find()` method on the array, to see if an element exists that matches your requirements.
* Thirdly, let's simply filter on whether the person has a "truthy" location, rather than comparing to the empty string. For example `undefined !== ''` but we don't want `location = undefined` to trigger the location column to appear.

I haven't tested this, but what I had in mind was something like the following:

```
let result = ['personDesc'];
if (this.people.find(p => p.location)) {
    result.push('location');
}
return result.concat(['availability', 'allocated', ...]);
```

src/app/group-summary/group-summary.component.css
```diff
@@ -2,3 +2,4 @@
 .viewoptions { margin: 0 0 16px; }
 .viewoptions mat-slide-toggle { margin-right: 16px; }
 .total app-resource-quantity { margin-left: 0.5em; }
+.total app-resource-quantity { margin-left: 0.5em; }
```
amdw, line 5: This looks like an accidental copy-paste?

src/app/people/people.component.html
```diff
@@ -7,6 +7,16 @@
     <td mat-cell *matCellDef="let element">{{element.personDesc}}</td>
     <td mat-footer-cell *matFooterCellDef>Total</td>
   </ng-container>
+  <ng-container *ngIf="showLocation" matColumnDef="personLocation">
```
amdw, line 10: I think you need to delete this one now, correct?

src/app/people/people.component.ts
```diff
@@ -50,16 +51,24 @@ export class PeopleComponent implements OnInit {
...
 
   ngOnInit() {
   }
 
+  displayedColumns(): string[] {
+    let columnLabels = ['personDesc'];
+    if (this.people.find(p => p.location)) {
+      columnLabels.push('location');
+    }
+    columnLabels = columnLabels.concat(['availability', 'allocated', 'unallocated', 'assignmentCount', 'commitFraction']);
```
amdw, line 66: Thinking about it further, I think you can just do this:

```suggestion
    columnLabels.push('availability', 'allocated', 'unallocated', 'assignmentCount', 'commitFraction');
```

src/app/period/period.component.html
```diff
@@ -55,6 +55,7 @@ <h4 class="mat-h4" [class.warning]="committedAllocationsTooHigh()">{{committedAl
     [totalAssignmentCount]="totalAssignmentCount()"
     [unit]="period.unit"
     [isEditingEnabled]="isEditingEnabled"
+    [showLocation]=true
```
amdw, line 58: This needs to be deleted now, right?

amdw:
Hey Sam - apologies if I jumped the gun, but I left some minor comments on some of your latest changes.

I'll hold off doing any more until you explicitly ask me to look at this again. There's a "re-request review" button at the top right (under "reviewers") which I believe will do the job, once you're ready.

Thanks!

Samrthi:
Hi Andrew,

Thanks for pointing the changes out, I must have accepted the wrong version at some merge conflict or accidentally used a wrong git command.
I’ve got the women engineers conference for three full days now but on Friday I will fix that and try to implement tests, so it’ll probably be only at the end of the day that I’ll re-request review.

Cheers,
Sam

amdw:
Thanks for letting me know Sam! Enjoy the conference - I hope you find it useful!

backend/main_test.go
```diff
@@ -80,7 +80,7 @@ func TestPostPeriod(t *testing.T) {
 
 	teamID := "myteam"
 	addTeam(handler, teamID, t)
-	periodJSON := `{"id":"2019q1","displayName":"2019Q1","unit":"person weeks","notesURL":"http://test","buckets":[{"displayName":"Bucket one","allocationPercentage":80,"objectives":[{"name":"Objective 1","resourceEstimate":0,"commitmentType":"Committed","notes":"some notes","assignments":[]}]}],"people":[]}`
+	periodJSON := `{"id":"2019q1","displayName":"2019Q1","unit":"person weeks","notesURL":"http://test","buckets":[{"displayName":"Bucket one","allocationPercentage":80,"objectives":[{"name":"Objective 1","resourceEstimate":0,"commitmentType":"Committed","notes":"some notes","assignments":[]}]}],"people":[{"ID": "alice", "DisplayName": "Alice Atkins", "Location": "LON", "Availability": 5}]}`
```
amdw, line 83: I was actually quite surprised this worked, since the keys you've used for the person in your JSON don't match the tags in the `struct` (the case is different, e.g. `ID` versus `id`).

Reading [the Go json docs](https://golang.org/pkg/encoding/json/#Unmarshal), this is apparently expected, since the unmarshalling prefers an exact match but will fall back to a case-insensitive match if one is available.

However, it would be better to use JSON which matches the kind the UI will produce. So please use JSON keys that exactly match the `json:` tags in the structs.

src/app/teamperiods/teamperiods.component.spec.ts
```diff
@@ -51,4 +52,11 @@ describe('TeamPeriodsComponent', () => {
   it('should create', () => {
     expect(component).toBeTruthy();
   });
+
+  it('should copy people list correctly', () => {
+    const people = [];
+    people.push(new Person('alice', 'Alice Atkins', 'LON', 5),
```
amdw, line 58: You can just use an array literal here, i.e. `const people = [new Person(...), new Person(...)]`. There is no need to start with an empty list and then push.

src/app/people/people.component.spec.ts
```diff
@@ -1,4 +1,4 @@
-// Copyright 2019 Google LLC
+// Copyright 2019-2020 Google LLC
```
amdw, line 1: If this file is no longer changing, then it shouldn't be part of this commit. Please could you revert it? Thanks!

PR #40, opened 2020-07-13
backend: refactoring. create models, storage, in_memory_storage packages

vavilen84:
Refactoring. Backend. Create new packages: models, storage, in_memory_storage.


vavilen84:
@googlebot I signed it!

vavilen84:
Hello, Andrew. My name is Vladimir. Nice to meet you :) Thank you for your response. I would like to participate in your project.

Firstly, I wanted to create docker environment for your project. It can make working with your project more comfortable in
cases when a contributor have many projects with different Go versions - instead of switching between different Go versions
globally contributor can just run docker. But when I tried to run backend with "--inmemstore" flag I got error:

Selection_289

this happened because "main" package is not a regular package in Go. This package is an entry point and all other logic
should be in a separate packages. Let me show you an example:
Selection_290
Selection_291
Selection_292
So, as we can see, despite the fact that both variables are placed in a single "main" package, b is undefined. I can’t call myself an expert in Golang, so please correct if I confuse something.

After I replaced undefined methods to a separate packages error disappeared (but you were right about "google_cds_store"
- I need to place it in a separate package also, because error is still here):
Selection_293

According to your notices:
1) license - I am going to do it.
2) default.go - Sometimes, if you have big package (several thousands strings of code) it is more developer friendly to split package into many go files. For example, if we have logic that relates exactly to each of models
it would make sense to create many files in "models" package, like "period.go","bucket.go" etc. (We have such logic now, but it is specific to "in memory storage", so it is okay not to replace this logic to abstract "models" package).  And if we have any common logic (or common variables/constants) for all models. so we could create a common file. I named such files "default.go" or "base.go". So in this case I choose this name. But at this stage I agree with you, these files can be named by the package name, we always have ability to rename them. So I am going to rename go files so they match theirs package names.
 3) No, I havn`t tried deploying this code to App Engine yet. I agree with, I am going to try it.

 According to git history - I agree with you. Such things like renaming, replacing, formatting, etc. make working with history more difficult.  But unfortunately, this is inevitable for projects that are being developed. Another thing is that perhaps you do not want
 to see this kind of refactoring in your own and you prefer to have logic in fewer packages. So, if you have other preferences
 of what you would like to see in your project - I am happy to take part and you only need to coordinate me in what to do.
 For example, we can defer this PR and I can do something else, so please, fill free to coordinate me.

amdw:
> Hello, Andrew. My name is Vladimir. Nice to meet you :) Thank you for your response. I would like to participate in your project.

Nice to meet you too! Thank you very much for your interest in PeopleMath, and for your reply.

> Firstly, I wanted to create docker environment for your project. It can make working with your project more comfortable in
> cases when a contributor have many projects with different Go versions - instead of switching between different Go versions
> globally contributor can just run docker. But when I tried to run backend with "--inmemstore" flag I got error:
> 
> Selection_289

I believe this happens because [the `go run` command](https://golang.org/cmd/go/#hdr-Compile_and_run_Go_program) compiles and runs a named *package*.

If you give it a single file, such as `go run main.go`, then the toolchain will try to treat that file as a package by itself. However, as you have observed, `main.go` is currently not a self-contained package, as it refers to names from the same package which are defined in other files.

You can make `go run` work against the current version by passing it the current directory (`.`) instead of `main.go`. This causes the toolchain to see all the files in the package:

```
$ go run . --inmemstore
2020/07/14 14:00:30 Using in-memory store per command-line flag
2020/07/14 14:00:30 Defaulting to port 8080
2020/07/14 14:00:30 Listening on port 8080
```

If this solution is sufficient for your needs, then great, and it's entirely up to you whether you continue with this pull request at all.

However, that doesn't mean your pull request is a bad idea. Splitting the code up into separate packages is a sound concept. If you'd like to continue to pursue it (with `google_cds_store.go` done as well) then that's great.

> 1. license - I am going to do it.

Thanks!

> 2. default.go - Sometimes, if you have big package (several thousands strings of code) it is more developer friendly to split package into many go files. For example, if we have logic that relates exactly to each of models
>    it would make sense to create many files in "models" package, like "period.go","bucket.go" etc. (We have such logic now, but it is specific to "in memory storage", so it is okay not to replace this logic to abstract "models" package).  And if we have any common logic (or common variables/constants) for all models. so we could create a common file. I named such files "default.go" or "base.go". So in this case I choose this name. But at this stage I agree with you, these files can be named by the package name, we always have ability to rename them. So I am going to rename go files so they match theirs package names.

I agree - thank you!

> 3. No, I havn`t tried deploying this code to App Engine yet. I agree with, I am going to try it.

If you've got time, then that's great, thanks. I'll be trying it myself either way. :) As I mentioned, I see no reason why it wouldn't work.

> 
> According to git history - I agree with you. Such things like renaming, replacing, formatting, etc. make working with history more difficult. But unfortunately, this is inevitable for projects that are being developed. Another thing is that perhaps you do not want
> to see this kind of refactoring in your own and you prefer to have logic in fewer packages. So, if you have other preferences
> of what you would like to see in your project - I am happy to take part and you only need to coordinate me in what to do.
> For example, we can defer this PR and I can do something else, so please, fill free to coordinate me.

That's very kind, thank you! The long-term health of the codebase has higher priority, and I agree splitting these packages makes sense to that end, so if you're interested in pursuing this PR then that's great, I'll accept it!

Given the `go run` suggestion above, I'd quite understand if you preferred to abandon it. It's entirely up to you.

Just one small warning: this change will have a small conflict with pull request #33, which is adding a `Location` field to the `Person` model. I expect that PR to be merged very soon (today or tomorrow). Whichever PR is ready first will get merged first, so the second person will have the (not-too-difficult) task of resolving the conflict! :)

Thanks again!

vavilen84:
Okay! Conflicts - not a problem :) I will resolve conflicts if need. After you merge your PR I can pull changes from master to my branch.

vavilen84:
Thanks for the explanation about "$ go run . --inmemstore" - I didnt know that.

vavilen84:
Fixes done:
- new google_cds_store package
- rename .go files for new packages according to package names
- add license headers to new .go files
- test Google App Engine deploy: https://peoplemath-v1.ue.r.appspot.com/
@amdw Please, check my fixes. Anything else I can do here?

amdw:
@vavilen84 - thank you very much!

It's interesting that you were able to successfully deploy to App Engine. After I wrote my last comment, I realised that with your changes, the `build_appengine.sh` script will no longer work, as [this line](https://github.com/google/peoplemath/blob/dbe57469d01d27ed0fc1e0cb43e086447edaf6bd/build_appengine.sh#L27) assumes the `backend` directory is flat.

How exactly did you create this App Engine deployment? Presumably you must have done something manual instead of (or after) running `build_appengine.sh`?

My commit 3ed46b6c3d227c4846d12ad7e86d608cb2a5131f should fix this problem, but I haven't had a chance to test it against your code yet. I'll do that tomorrow.

In the meantime, the other pull request #33 has been merged, so there are now some conflicts with your changes. Would you be able to resolve those and merge the latest changes from master? I'll test an App Engine deploy of my own sometime tomorrow.

Thanks again!

vavilen84:
@amdw You was right - I changed  build_appengine.sh but forgot about it and reverted:
Selection_298
thanks for your fix! I resolved conficts, test passed successfully (go & frontend) and gcloud deploy also works fine with these changes. So, please, review merge.

amdw:
@vavilen84 - many thanks! I was able to deploy to App Engine using your latest changes, so I think we're very close to being able to merge this now.

One thing I just noticed is that the `CommitmentTypeAspirational` and `CommitmentTypeCommitted` constants have been left behind in `main.go` - these really belong with the model types in `models.go`.

Would you be able to move them there, keeping them in the same place they originally were (just before the `Assignment` type)? (I'd like to be able to run a `diff` between the old `main.go` and the new `models.go` to clearly see that the moved code is the same before and after.)

Don't forget to update the references in `main.go` (the tests will remind you if you forget!).

Once you've done that, please re-request a review and I'll do the merge.

vavilen84:
@amdw Done: CommitmentTypeAspirational and CommitmentTypeCommitted constants replaced to models.go before Assignment type.

amdw:
Thank you very much for contributing to PeopleMath!

vavilen84:
@amdw Hi! Sorry for offtopic - I have an idea for my next refactoring. 
- use https://github.com/gorilla/mux - this is flexible tool, this will allow us to create a separate route for each method (GET, POST) so we dont need to use if/else condition. 
- get rid of Server struct - for now this entity contains store and defaultTimeout. defaultTimeout - I suggest to put it in "storage" package. StorageService can be placed globally as a variable in "storage" package. This improved should give us "low-coupling", because storage and REST API actions are different abstractions in a programm.
- create new package "controllers" and create 2 controllers files in it: team.go period.go. Inside controllers we can put handlers like Get() Post() Put() - i think it could be cool.
What do you think about it?

amdw:
Hi @vavilen84 - thank you for your suggestions!

As you say, they're off-topic for this pull request, so I'd suggest you open an issue for each one (I think you should have permissions to do that?) and we can discuss them in more detail there.

My initial reactions:

* Sounds good - the Gorilla libraries seem quite well-regarded, so if this simplifies the code then I'm all for it!
* I'm open to suggestions on this, but I'm not generally a big fan of global variables. The fact that `StorageService` is behind an interface is already intended to provide loose coupling between the HTTP server and the storage implementation, and the `storeTimeout` (which determines how long the server is prepared to wait for the store to return a value) seems more a concern of the HTTP server than the storage service. Having said that, I'm happy to look at what you have in mind.
* Splitting the team-related and period-related controllers into separate files sounds reasonable to me if we can make it work. There may be some code shared between those controllers, but we can cross that bridge when we come to it.

Regarding the choice of data store: a NoSQL database seemed like a good fit because the `Period` data structure is quite deeply nested (not having to do joins potentially brings some performance gains), and we have quite simple querying needs. The lack of schema makes it very easy to add new fields - you just modify the Go structs and everything just works.

The nice thing about [Cloud Datastore](https://cloud.google.com/datastore) specifically is that there is quite a generous free tier, meaning that you can run small instances on App Engine without paying anything, which is not the case even for the smallest SQL database (as [Cloud SQL](https://cloud.google.com/sql/docs) instances require a dedicated VM).

That said, I can well imagine that some users would prefer to use a SQL data store, and I'm quite happy to accept a new version of the `StorageService` that uses a SQL database instead, perhaps using [Gorm](https://gorm.io/) or similar. The only thing is that I'm not likely to use it myself, so I can't promise to invest time in maintaining it.

PR #42, opened 2020-07-16
add docker dev environment

vavilen84:
@amdw Hi! This PR add an alternate way to run project - via docker-compose. Appropriate readme instructions are added. Please, review.

### DIFF
--- a/.env.dist
+++ b/.env.dist
@@ -1,0 +1,1 @@
+WORKING_DIR=/path/to/project
--- a/.env.dist
+++ b/.env.dist
@@ -1,1 +1,0 @@
-WORKING_DIR=/path/to/project
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,5 @@ testem.log
 # System Files
 .DS_Store
 Thumbs.db
+
+.env
--- a/.gitignore
+++ b/.gitignore
@@ -41,5 +41,3 @@ testem.log
 # System Files
 .DS_Store
 Thumbs.db
-
-.env
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,5 @@
+FROM node:lts
+
+RUN npm install -g @angular/cli @angular-devkit/build-angular
+RUN npm install
+CMD ng serve --port 4200 --host 0.0.0.0 --disable-host-check
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,0 @@
-FROM node:lts
-
-RUN npm install -g @angular/cli @angular-devkit/build-angular
-RUN npm install
-CMD ng serve --port 4200 --host 0.0.0.0 --disable-host-check
--- a/README.md
+++ b/README.md
@@ -61,6 +61,33 @@ To persist the API data and exercise the Cloud Datastore persistence layer, the 
 
 The front-end tests can be run via `ng test`, and the back-end tests via `go test` in the `backend` directory.
 
+## Docker dev environment
+
+This is an alternative way to run project in dev environment.
+
+[Install Docker](https://docs.docker.com/get-docker/)
+
+[Install docker-compose](https://docs.docker.com/compose/install/)
+
+Rename .env.dist -> .env
+
+Set correct WORKING_DIR env variable in .env (path to root folder of project)
+
+Run project
+```
+$ cd /path/to/project
+$ docker-compose up
+```
+
+Project should be available via the link http://localhost:4200
+
+Stop project
+```
+$ cd /path/to/project
+$ docker-compose down
+```
+
+
 ## Deployment to App Engine
 
 To deploy the app to Google App Engine:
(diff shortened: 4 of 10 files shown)


amdw:
Hi @vavilen84 - thanks for your pull request.

Unfortunately I don't think I can accept your proposed change to `devproxy.conf.json`, because it will break the development workflow outside Docker (the hostname `backend` won't be resolvable).

I'm open to the possibility of adding configuration for `docker-compose`, but the regular development workflow must continue to work without any complications or extra steps.

Could we perhaps put all the Docker-related configuration (`Dockerfile`s, `docker-compose.yaml` etc) in a special folder, maybe called something like `docker_dev`, and then have a script called something like `docker_dev/setup.sh` which creates the `.env` file, patches in any necessary changes to `devproxy.conf.json`, and creates symlinks to the other files wherever they are needed? This would save the user from manually setting up the `.env` file, for example, and it would keep all the Docker-related files in one place so that people who weren't using Docker wouldn't be confused as to why they are there.

In that case we should change `.gitignore` to ensure that Git still sees the repository as clean after `docker_dev/setup.sh` has been run, though if it really isn't possible to make this work without changes to `devproxy.conf.json`, I don't see how we can avoid that file showing up as changed after the script has been run, which would be unfortunate.

The README doesn't explain why someone might prefer to use a Docker Compose development workflow over the currently suggested one. I think we would need to add some explanation of this, as I must confess I'm a little unclear on that myself at the moment.

Virtually all the dependencies the project has are hermetically captured in `package.json` (for JavaScript) and `backend/go.mod` (for Go), so the only additional ones you'd capture by using Docker Compose are the versions of Node and Go themselves (plus any Linux libraries, binaries etc these transitively depend on). But these can easily be managed via version managers like [nvm](https://github.com/nvm-sh/nvm) for Node. So what benefits does Docker Compose give us here?

On a related topic, why are you running `apt-get update` in the `Dockerfile` for the backend? Surely this just means you're getting whatever version of Go is currently the latest in the `apt` repository of the Linux distribution the base image is based on? But isn't that exactly what you'd get by simply running the project on a regular Linux system? Again, what benefit is Docker Compose bringing us here?

So in summary:

* Please ensure that the non-Docker development workflow continues to work correctly. I believe this rules out changes like the one you've made to `devproxy.conf.json`.
* Please consider my idea of putting all the Docker-related configuration files in a special folder and adding a setup script that configures the workspace for use with Docker Compose.
* Please add an explanation to the README of why someone might prefer a Docker Compose workflow over the current workflow.

Thank you!

vavilen84:
@amdw Thanks for your feedback - I completelly agree with you.

 devproxy.conf.json - I forgot that changing this will break working without docker env. I like your idea about placing all related docker files in one place and just have one console command to init docker/regular env (to switch between environments)

Why do we need docker? I do not consider the use of this technology as something necessary. But there is a possibility, when versions of go & node which are used by contributor are different from what are used in your project. In this case it would make difficulties for contributor (Ex.: I work on a project which use Go 1.11 and I want to contribute to your project. So I need to switch between different Go versions In order to work on both projects. Docker just simplify this process - just run from console "docker-compose up" and you work in isolated environment). But if you think that this is not required exactly for your project - we can just abandon this PR. If you decide to move forward here - I will add more detailed description when it can be used by a contributor. 

"RUN apt-get -y update" - thank you for your notice, I left it by mistake. But if we add 'apt-get upgrade' - it will not change go version, go was compiled from source.

amdw:
@vavilen84 - many thanks for explaining! I wasn't intending to disparage Docker Compose, just to better understand what you think its benefits would be in the context of this project, especially compared to version managers like `nvm`.

I'm happy to accept a PR facilitating the use of Docker Compose, provided:

* it doesn't break the existing non-Docker-based development workflow,
* any files specific to this workflow (such as `Dockerfile`s or `docker-compose.yaml`) don't distract or confuse people who aren't using it, and
* the README links to documentation explaining what Docker Compose is and what its benefits are ([this page](https://docs.docker.com/compose/) seems like a reasonable choice).

Given that your frontend container currently runs the Angular CLI, which isn't suitable as a production web server, I think it's also important that the README makes clear that the Docker Compose configuration shouldn't be used in production.

The Docker Compose documentation says "Compose works in all environments: production, staging, development, testing, as well as CI workflows", but the configuration you're writing seems suitable only for the development phase. This is fine for now, but I wouldn't be surprised if someone comes along wanting to run the application this way in production at some stage. At that point we might have to change the configuration to something that can work in production.

Many thanks!

vavilen84:
@amdw Thanks for your feedback - changes are done:
- All docker related files moved to a separate /docker/dev folder (in future we can create separate folder /docker/production for production environment if we need)
- README.md updated
- src/devproxy.conf.json fixed
Please, review changes.

amdw:
Thanks very much! I'll make a few changes of my own to the README shortly.

PR #51, opened 2020-07-29
add gorilla mux

vavilen84:
@amdw Hi! Please review. This PR should add Gorilla MUX and use mux.Vars(r) functionality in handlers. Related PR https://github.com/google/peoplemath/pull/49.

### DIFF
--- a/backend/go.mod
+++ b/backend/go.mod
@@ -5,6 +5,7 @@ require (
 	cloud.google.com/go/datastore v1.1.0
 	github.com/golang/protobuf v1.3.4 // indirect
 	github.com/google/uuid v1.1.1
+	github.com/gorilla/mux v1.7.4
 	golang.org/x/exp v0.0.0-20200228211341-fcea875c7e85 // indirect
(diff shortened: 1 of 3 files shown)


amdw:
Thank you very much for sending this pull request! Apologies that it's taken me a week to get to it.

In terms of its intent, this is perfect - a simple conversion to Gorilla Mux as we discussed in #49.

Unfortunately, I think the current proposed change would break a few HTTP endpoints. For example, `GET /api/team/`, to load a list of all teams, no longer works with these changes. I suspect `/api/team/{teamID}` does not match the case where `teamID` is empty.

This is largely my fault, because the `main_test.go` was missing coverage of several HTTP endpoints. I just pushed commit 8a29ed30bfb7e51ae479e2c6c397b12657cdfb31 to fix that. These new tests pass on `master` but fail against this PR.

Please could you fix your pull request so that these tests pass? Many thanks!

vavilen84:
Thank you for pushing tests - I fixed routers according to them & divided GET endpoints according to new routers. Please, review.

Issue #39, opened 2020-07-10
Make HTTP errors more noticeable


amdw:
A couple of times I've accidentally dismissed the snackbar that tells you a concurrent edit has been made.

This error should be made more noticeable - perhaps a modal that can only be dismissed by clicking a certain button, and/or maybe pressing the escape key?

It should prompt the user to copy any unsaved work and then reload the page.

amdw:
This is probably an issue with all errors, not just concurrent modification errors.

amdw:
This should now be much easier to resolve using the new `NotificationService` introduced by @Samrthi.

r-sreesaran:
I would like to work on this issue.

amdw:
@r-sreesaran - great, thanks very much!

All these errors should go through the `NotificationService`, so that we can decide centrally there how to display them. Off the top of my head I'm not sure how consistently we're already doing that.

PR #52, opened 2020-08-05
Staging

Samrthi:


src/app/services/auth.service.ts
```diff
@@ -0,0 +1,72 @@
...
+  constructor(
+    private router: Router,
+    private notificationService: NotificationService,
+    public angularFireAuth: AngularFireAuth
+  ) {
+    if (environment.requireAuth) {
+        angularFireAuth.onAuthStateChanged(firebaseUser => {
+          if (firebaseUser != null) {
+            const user: User = {uid: firebaseUser.uid, displayName: firebaseUser.displayName};
+            this.user$ = of(user);
```
amdw, line 27: I'm a bit surprised that we'd create a brand-new `Observable` every time the user changes. I thought the point of an `Observable` was that it provides a stream of values, to which other components can then subscribe?

Presumably in this case, the intent is to enable other components to subscribe for who the current user is?

In that case I'd have thought we'd make `user$` a `readonly` member, and then initialize it in the constructor to something like a `BehaviorSubject`:

https://rxjs.dev/guide/subject#behaviorsubject

Then, you call `next()` on that subject with the new value inside your `onAuthStateChanged` handler?

This way, `user$` is a constant `Observable` which emits a new value every time the user changes.

src/app/services/auth.service.ts
```diff
@@ -0,0 +1,72 @@
...
+          } else {
+            this.user$ = of(null);
+          }
+        });
+      }
+  }
+
+  googleSignin(): void {
+    const provider = new auth.GoogleAuthProvider();
+    this.angularFireAuth.signInWithPopup(provider).then(result => {
```
amdw, line 37: Is there a type annotation you can add for `result` here?

src/app/services/auth.service.ts
```diff
@@ -0,0 +1,72 @@
...
+      const user = result.user;
+      if (user == null) {
+        throw new Error('User is null');
```
amdw, line 40: Under what circumstances will this condition happen in practice? And what is the user-visible behaviour when you throw an `Error` like this?

src/app/services/auth.service.ts
```diff
@@ -0,0 +1,72 @@
...
+      } else {
+        this.notificationService.notification$.next('Signed in as ' + user.displayName);
+        this.router.navigate(['/']);
+        return this.updateUserData(user);
+      }
+    }).catch(error => {
+      console.log(error);
```
amdw, line 47: Maybe we should send this to your notification service as an error, so we can make a central decision about how we want to display errors?

src/app/services/auth.service.ts
```diff
@@ -0,0 +1,72 @@
...
+  public getIdToken(): Observable<string | null> {
+    return this.angularFireAuth.idToken;
+  }
+
+  public signOut(): void {
+    this.angularFireAuth.signOut().then(result => {
+      this.user$ = of(null);
+      this.router.navigate(['/login']);
+    }).catch(error => {
+      console.log('Sign out error: ', error);
```
amdw, line 69: As above... this should probably go via the notification service.

src/environments/firebaseConfig.ts
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright 2020 Google LLC
```
amdw, line 2: Can you add the full Apache license header here?

The `addlicense` tool is the easiest way to do this, [as described here](https://opensource.google/docs/releasing/preparing/#license-headers).

On a Unix-like system, you can automatically add any missing headers to TypeScript files with a shell command like:

`find . -name \*.ts | xargs addlicense -c "Google LLC" -l apache`

This shouldn't touch any files that already have the right header.

src/app/app.module.ts
```diff
@@ -88,10 +96,23 @@ import { PillComponent } from './pill/pill.component';
     MaterialModule,
     FlexLayoutModule,
     HttpClientModule,
+    AngularFireModule.initializeApp(firebaseConfig.firebase, 'firebaseApp'),
```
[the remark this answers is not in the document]
Samrthi: I used it for debugging when I had problems with the initialization and then forgot to change it back - changed it now since it doesn't have any benefit.

src/app/app.component.ts
```diff
@@ -13,12 +13,25 @@
...
 
 @Component({
   selector: 'app-root',
   templateUrl: './app.component.html',
   styleUrls: ['./app.component.css']
 })
 export class AppComponent {
-  title = 'PeopleMath'
+  constructor(
+    public auth: AuthService,
```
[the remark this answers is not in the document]
Samrthi: I checked and yes, otherwise it will throw an error because it is accessed from app.component.html.

src/app/models/user.model.ts
```diff
@@ -0,0 +1,7 @@
+/*
+ * Copyright 2020 Google LLC
+ */
+export interface User {
+  uid: string | undefined;
```
amdw, line 5: Can we not just say `uid?: string` here?

src/app/models/user.model.ts
```diff
@@ -0,0 +1,7 @@
...
+  displayName?: string | null;
```
amdw, line 6: Why do we need ` | null`? We already have `?:` here which allows `displayName` to be `undefined`. Not sure I see we need to make the user deal with both `undefined` and `null` here.

src/app/app.component.css
```diff
@@ -4,3 +4,7 @@
 @media screen {
     .mat-sidenav-container { height: 100% }
 }
+
+button {
+  margin: 0 10px;
```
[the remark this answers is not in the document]
amdw: Not sure I got an answer to this one? :)

src/app/services/auth.service.ts
amdw: Not sure I got an answer to this one?

src/app/services/auth.service.ts
amdw: Not sure I got an answer to this one? Apologies if we discussed it and I forgot, but it would be good to note the answer here for the record.

src/app/services/auth.service.ts
amdw: At the very least, I think these should use `console.error` so they show up red in the console.

src/app/services/auth.service.ts
amdw: Again, this should be `console.error` at minimum.

src/app/services/auth.service.ts
Samrthi: Ah my bad, yes, I added it.

src/app/models/user.model.ts
Samrthi: I added it because it can come back as null from Firebase but thinking about it more it makes more sense to just check if it returns null and only add it if it does exist, I'll implement that.

src/app/models/user.model.ts
Samrthi: Yep, good point. (Still new to the ? operator so I missed it)

src/app/services/auth.service.ts
Samrthi: I changed it to
`this.notificationService.error$.next(error);`
in my last push, not sure if you saw that, I assume that is appropriate?

src/app/services/auth.service.ts
Samrthi: As above.

src/app/app.component.css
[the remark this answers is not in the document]
Samrthi: I need to remember to actually reply to you, sorry!
Yes, you're right. I changed it to use the already existing id on the button now instead.

src/app/services/auth.service.ts
Samrthi: We did discuss this, I implemented it this way because I wasn't aware I could write `user?.displayName`. I've changed the implementation now so it doesn't throw an error here.

src/app/services/auth.service.ts
amdw: My mistake, sorry - I assumed because this code snippet was still here in the review, that this code was still on the head of your branch.

src/app/services/auth.service.ts
amdw: As above - this was my mistake, sorry.

src/app/app.component.ts
```diff
@@ -13,12 +13,25 @@
...
+    private notificationService: NotificationService,
```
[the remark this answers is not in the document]
amdw: Leaving this unresolved for now as a reminder - let's close it when we've updated #39.

Samrthi:
`import { Subject } from 'rxjs';` does work to get rid of one of the warnings but I wasn't able to find alternative imports for firebase. I followed the imports a given in this example: https://github.com/angular/angularfire/blob/HEAD/docs/auth/getting-started.md
 and I can't seem to find any alternative imports that ECMAScript modules. The only 'solution' to these warning (specifically for firebase) I have found online was to add them to the `allowedCommonJsDependencies ` as described on the webpage in the warning.

amdw:
> The only 'solution' to these warning (specifically for firebase) I have found online was to add them to the `allowedCommonJsDependencies ` as described on the webpage in the warning.

I think the problematic import might be this one: `import * as firebase from 'firebase';`

It's a bit annoying, because you're only using it to do type-checking on the user. You don't actually seem to be using those imports at runtime at all.

If we really wanted to get rid of this, we could probably just replace `firebase.User` in your `AuthService` with an `interface` we define, maybe called something like `FirebaseUser`, that simply lists the fields we want to access.

However, it's probably OK to just add `firebase` to the `allowedCommonJsDependencies`.

In any event, it's important that we don't get any of these warnings when we build the app. Thanks!

Samrthi:
I added these three:
`"allowedCommonJsDependencies": [
              "firebase",
              "@firebase/app",
              "@firebase/auth"
            ]`
Each of them had one warning and I also didn't seem to be able to just do `"@firebase` instead of the two separate ones. I also changed the import for the user model to
`import {User as firebaseUserModel} from 'firebase';`
The `firebase` import is also used here `import {auth} from 'firebase/app'` which caused the same warning even when removing the other import statement.
Let me know if it is okay this way.

amdw:
Thanks Sam - these warnings are gone from the build now, and I think this is all we can do for the moment.

I tried running the Angular tests against your code, and I saw quite a few warning messages that don't happen against `master`. For example I see quite a few like this:

```
ERROR: ''mat-card' is not a known element:
1. If 'mat-card' is an Angular component, then verify that it is part of this module.
2. If 'mat-card' is a Web Component then add 'CUSTOM_ELEMENTS_SCHEMA' to the '@NgModule.schemas' of this component to suppress this message.'
```

Normally this means you need to add `MaterialModule` to the `imports` section of your `TestBed` config.

Please could you eliminate these error messages before we submit this change? Thanks!

The following warnings are expected and occur on `master`:

```
WARN: 'Spec 'ImmutablePeriod should be immutable' has no expectations.'
WARN: 'Spec 'ImmutableBucket should be immutable' has no expectations.'
WARN: 'Spec 'ImmutableObjective should be immutable' has no expectations.'
```

PR #54, opened 2020-08-07
Move handlers from 'main' package

vavilen84:
Related PR https://github.com/google/peoplemath/pull/49. Changes:
- New 'server' package with handlers
- Fix test according to new changes
- Fix GET improve method - remove redunant method validation since we have implemented Gorilla Mux
@amdw Please, review changes. Thanks.


backend/server/router.go
```diff
@@ -0,0 +1,38 @@
...
+	"time"
+)
+
+// Server struct to handle incoming HTTP requests
+type Server struct {
+	store        storage.StorageService
+	storeTimeout time.Duration
+}
+
+type Config struct {
```
amdw, line 16: Why introduce a new `Config` type which has exactly the same members as `Server`, and then copy over those members one-by-one in `MakeHandler`? Why not just create the `Server` type directly?

My natural instinct here is that we should continue to instantiate the `Server` type as we are doing today (either via a struct literal or a static factory function) and keep `MakeHandler` as an instance method on `*Server`. Can you see any reason not to do that?

vavilen84: My idea was to keep fields of Server struct not exported and make server configuring more explicit. But I think that you are right - we don't need it for now. I will refactor this.

backend/server/router.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+func InitServer(store storage.StorageService, storeTimeout time.Duration) *Server {
+	return &Server{
+		store:        store,
+		storeTimeout: storeTimeout,
+	}
+}
+
+func MakeHandler(s *Server) http.Handler {
```
amdw, line 27: Any reason we can't just make this a method on the server, like it was originally?

`func (s *Server) MakeHandler() {...`

amdw (changes requested):
Really sorry for my slow response on this. Thank you for your latest changes.

A couple of minor comments inline. Also, there are some diffs showing up here (e.g. in the tests) from recent commits on `master` - ideally these should be merged so that they don't show up in the diff.

Also, just a warning that PRs #56 and #57 will be merged soon, and will cause conflicts with this change. If you like, we can just leave this PR until after that's done, and I can make these changes myself after that's happened to save you the conflict resolution effort? They're fairly simple and mechanical.

Thanks again for your contributions!

backend/server/router.go
vavilen84: Hi, sorry for the slow response on this! This method is just for separate routes from main.go. Do you have an idea to leave creating handlers in main.go? If so - it would not be a problem for me to move them back where they were initially.

vavilen84:
Changes from master merged.

vavilen84:
@amdw Hey, changes from master are merged and conflicts are resolved - you can review now.

amdw:
Thank you for these latest commits, and apologies once again for the delay in replying.

I intend to merge #61 before this PR, which will likely cause some more conflicts, unfortunately. I will post an update with next steps once I have one.

amdw:
Hello! This PR has conflicts again, and I felt bad about constantly asking you to merge in conflicting changes from the master branch, so I went ahead and did this refactoring myself and merged it in c1a36b84a71d113f38fd259d9a37f2251a218426.

Given that commit has now happened, are you happy to close this PR?

I apologise for the amount of time it has taken to get this resolved. Thanks!

vavilen84:
@amdw Hi! Thank you! Yes, sure, lets close it.

stuti-ui:
Hi.
Yeah sure let's close it.
Thank You

PR #56, opened 2020-08-18
Backend authentication

Samrthi:


README.md
```diff
@@ -87,3 +87,11 @@ The primary supported way of running PeopleMath in a staging or production confi
 Authentication using Firebase is in development.
 Currently, only the frontend is implemented. As long as
```
amdw, line 88: This line is no longer true as of this PR. :) Can you re-check this whole section?

backend/main.go
```diff
@@ -42,20 +45,61 @@ const (
 type Server struct {
 	store        storage.StorageService
 	storeTimeout time.Duration
+	auth         Auth
+}
+
+type Auth interface {
```
amdw, line 51: These auth-related types would probably be best put in a new package, called `auth` or something.

backend/main.go
```diff
@@ -42,20 +45,61 @@ const (
...
+
+func (auth noAuth) authenticate(next http.HandlerFunc) http.HandlerFunc {
+	return next
+}
+
+type firebaseAuth struct {
+	firebaseClient authClient
+}
+
+type authClient interface {
```
amdw, line 65: I believe this is the interface you're writing specifically to match our expectations of the Firebase client, so we can mock it out in tests, correct?

If so then it should probably have "Firebase" somewhere in its name. If you're moving these types to another package, we should probably have a file in there called `firebase.go` which contains all the Firebase-specific code.

backend/main.go
```diff
@@ -42,20 +45,61 @@ const (
...
+	VerifyIDToken(ctx context.Context, idToken string) (*auth.Token, error)
+}
+
+func (auth firebaseAuth) authenticate(next http.HandlerFunc) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		ctx, cancel := context.WithTimeout(r.Context(), defaultStoreTimeout)
```
amdw, line 71: We shouldn't use `defaultStoreTimeout` here. This was only intended for the storage (and can be overridden by the user anyway).

We should have a separate timeout on the `Server` struct, called `authTimeout` or something like that.

The value of `defaultStoreTimeout` may be an equally reasonable default for both the storage and the auth, in which case it would be fine to rename that constant.

backend/main.go
```diff
@@ -42,20 +45,61 @@ const (
...
+			log.Printf("User authentication failed: error: %s", err)
+			http.Error(w, "User authentication failed (see server log)", http.StatusUnauthorized)
+			return
+		}
+
+		next(w, r)
+	}
+}
+
+func getDomain(email string) string {
```
amdw, line 86: I think this is part of your later authorization-related change, right? I don't see it called anywhere here. It probably shouldn't be part of this pull request.

backend/main.go
```diff
@@ -352,7 +398,29 @@ func main() {
 			return
 		}
 	}
+
 	server := Server{store: store, storeTimeout: defaultStoreTimeout}
```
amdw, line 402: It's a minor thing, but can we instantiate the `Server` *underneath* this block of code, and put our `auth` into a `var`, same as we do above for the `store`?

This is just a bit more consistent: we gather all the things we need to pass into the server and then we instantiate it - rather than instantiating the `Server` with half the things it needs and then setting the others later.

backend/main.go
```diff
@@ -352,7 +398,29 @@ func main() {
...
+		}
+		firebaseClient, err := app.Auth(ctx)
+		if err != nil {
+			log.Fatalf("Could not get Firebase Auth client: %v\n", err)
+			return
+		}
+
+		firebaseAuth := firebaseAuth{firebaseClient: firebaseClient}
+		server.auth = firebaseAuth
+	}
```
amdw, line 422: We should have an `else` here that calls `log.Fatalf` to tell the user they requested an unsupported `authMode`.

backend/main_test.go
```diff
@@ -27,6 +31,12 @@ import (
 	"testing"
 )
 
+func makeHandler() http.Handler {
+	server := Server{store: in_memory_storage.MakeInMemStore()}
+	server.auth = noAuth{}
```
amdw, line 36: Again, a nit: can you just pass this directly in when you instantiate the `Server` struct? I.e.

`server := Server{store: ..., auth: ...}`

backend/main_test.go
```diff
@@ -476,3 +465,80 @@ func TestImproveBadMethods(t *testing.T) {
...
+	} else {
+		token := &auth.Token{}
+		err := errors.New("token invalid")
+		return token, err
+	}
+}
+
+func TestFirebaseAuth(t *testing.T) {
+	server := Server{store: in_memory_storage.MakeInMemStore()}
+	server.auth = firebaseAuth{firebaseClient: AuthClientStub{}}
```
amdw, line 529: Again, this could be  inline, e.g. `auth := firebaseAuth{...}` and then `server := Server{store: ..., auth: auth}`

backend/main_test.go
```diff
@@ -476,3 +465,80 @@ func TestImproveBadMethods(t *testing.T) {
 	postresp := makeHTTPRequest(postreq, handler, t)
 	checkResponseStatus(http.StatusMethodNotAllowed, postresp, t)
 }
+
+type failingAuthStub struct{}
```
amdw, line 469: Nit: we should probably be explicit that it's *authentication* (rather than authorization) that's failing here. `failAuthenticationStub` maybe?

backend/main_test.go
```diff
@@ -476,3 +465,80 @@ func TestImproveBadMethods(t *testing.T) {
...
+
+func (auth failingAuthStub) authenticate(next http.HandlerFunc) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		http.Error(w, "Auth failed", http.StatusUnauthorized)
+		return
+	}
+}
+
+func TestAuthMiddleware(t *testing.T) {
+	server := Server{store: in_memory_storage.MakeInMemStore()}
```
amdw, line 479: Nit: `..., auth: failingAuthStub{}}`

backend/main_test.go
```diff
@@ -476,3 +465,80 @@ func TestImproveBadMethods(t *testing.T) {
...
+	server.auth = failingAuthStub{}
+	handler := server.makeHandler()
+
+	testHandlerWithAuth := func(httpMethod, target string, body io.Reader) {
```
amdw, line 483: Ideally the name would be a bit clearer as to what this function is actually doing. What it's doing is calling the supplied HTTP endpoint and asserting that we get a 401 error.

`assertAuthenticationFailure` maybe?

backend/main_test.go
```diff
@@ -476,3 +465,80 @@ func TestImproveBadMethods(t *testing.T) {
...
+		req := httptest.NewRequest(httpMethod, target, body)
```
amdw, line 484: Do you ever pass anything non-`nil` as the `body` here? If so, does the function really need this parameter?

backend/main_test.go
```diff
@@ -476,3 +465,80 @@ func TestImproveBadMethods(t *testing.T) {
...
+	testHandlerWithAuth(http.MethodGet, "/api/team/"+teamID, nil)
+	testHandlerWithAuth(http.MethodGet, "/api/team/", nil)
+	testHandlerWithAuth(http.MethodPost, "/api/team/", nil)
+	testHandlerWithAuth(http.MethodPut, "/api/team/"+teamID, nil)
+	testHandlerWithAuth(http.MethodGet, "/api/period/"+teamID+"/"+periodID, nil)
+	testHandlerWithAuth(http.MethodGet, "/api/period/"+teamID+"/", nil)
+	testHandlerWithAuth(http.MethodPost, "/api/period/"+teamID+"/", nil)
+	testHandlerWithAuth(http.MethodPut, "/api/period/"+teamID+"/"+periodID, nil)
+
+	getreq := httptest.NewRequest(http.MethodGet, "/improve", nil)
```
amdw, line 500: This could probably use a comment explaining that you're asserting here that `/improve` is not covered by authentication.

Either that or you could write another helper function called something like `assertNoAuthenticationFailure`, that makes the opposite assertion to the one above. You could either pass in the specific status code expected, or just assert that you get something in the `200` or `300` range.

backend/main_test.go
```diff
@@ -476,3 +465,80 @@ func TestImproveBadMethods(t *testing.T) {
...
+	checkResponseStatus(http.StatusFound, getresp, t)
+}
+
+type AuthClientStub struct{}
+
+func (AuthClientStub) VerifyIDToken(ctx context.Context, idToken string) (*auth.Token, error) {
+	claims := map[string]interface{}{
+		"user_id":        "id123",
+		"email_verified": true,
+		"email":          "test@google.com"}
```
amdw, line 511: Nit: closing `}` on the next line here, with a trailing comma so we don't have to modify the `email` line if we want to add more elements to this struct?

README.md
```diff
@@ -87,3 +87,11 @@ The primary supported way of running PeopleMath in a staging or production confi
...
 `requireAuth` is set to `false` in `src/environments/environment.prod.ts` the app will run as expected without authentication.
+
+To enable authentication using Firebase:
+
+* Set `requireAuth` to `true` in `src/environments/environment.prod.ts`
+* Download your [firebase config file](https://support.google.com/firebase/answer/7015592#web) and copy the text into the `firebaseConfig.ts` file in the same folder.
```
amdw, line 94: Nit: `Firebase` (capital F)

Same on the line below.

amdw (changes requested):

backend/main.go
Samrthi: Yep, that's true. I've removed it.

backend/main.go
Samrthi: I've put a default auth timeout into the auth package.

README.md
Samrthi: Resolved.

backend/main.go
Samrthi: Created package auth with auth.go and firebase.go

backend/main.go
Samrthi: Renamed to firebaseAuthClient

backend/main.go
Samrthi: Makes sense, done.

backend/main.go
Samrthi: Good point, done.

README.md
Samrthi: I've edited now, not sure about the flag, did we say we'd rather have this as a setting? I skimmed the settings implementation and couldn't find an easy way to implement that.

backend/main_test.go
Samrthi: Yep, makes perfect sense, don't know why that didn't cross my mind 😅

backend/main_test.go
Samrthi: 👍🏻

backend/main_test.go
Samrthi: Done

backend/main_test.go
Samrthi: It doesn't, I've removed it.

backend/main_test.go
Samrthi: Done

backend/main_test.go
Samrthi: Yes, I definitely struggled to find an expressive, readable name for this, I've changed it to your suggestion.

backend/main_test.go
Samrthi: I added a comment

backend/main_test.go
Samrthi: I think the trailing comma threw me off a bit, I changed it now.

README.md
amdw: For now I think it is OK for it to be a flag. As and when I roll this out to our internal instance, I may move it to the `Settings`. The way to implement it would be to load the settings from the storage, after the `StorageService` has been instantiated but before the `Server` has.

amdw (approved):
Thanks for making the changes. If you can fix the CLA check then I think we're good to merge this.

Samrthi:
New clean pull request is opened

PR #60, opened 2020-08-26
Basic authorization using email domain

Samrthi:
I implemented the domain as a flag, but @amdw if you would prefer to implement it as a setting somehow, I'm happy to change the implementation if you can give me a few more details on how you think settings would ideally be implemented. Just looking at the settings you implemented I could not see an intuitive way to use the existing settings implementation for auth.


backend/auth/firebase.go
```diff
@@ -15,36 +15,56 @@
...
 
 type firebaseAuthClient interface {
 	VerifyIDToken(ctx context.Context, idToken string) (*auth.Token, error)
 }
 
-func (auth FirebaseAuth) Authenticate(next http.HandlerFunc) http.HandlerFunc {
+func (auth FirebaseAuth) Authorize(next http.HandlerFunc) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
-		ctx, cancel := context.WithTimeout(r.Context(), auth.AuthTimeout)
+		_, cancel := context.WithTimeout(r.Context(), auth.AuthTimeout)
```
amdw, line 37: This line doesn't do anything any more. I think it needs to be moved into `Authenticate`. You will need to pass `r.Context()` into it. See below.

backend/auth/firebase.go
```diff
@@ -15,36 +15,56 @@
...
+		if getDomain(userEmail) != *auth.AuthDomain {
+			http.Error(w, "You are not authorized to view this resource", http.StatusForbidden)
+			return
+		}
 		next(w, r)
 	}
 }
+
+func (auth FirebaseAuth) Authenticate(idToken string) (userEmail string, httpError *string) {
+	token, err := auth.FirebaseClient.VerifyIDToken(context.Background(), idToken)
```
amdw, line 57: `context.Background()` is not correct here. We need to use the context of the HTTP request (and add a timeout with deferred `cancel()`, like we were doing before).

You will need to pass that context in as a parameter to this function. By convention in Go, this is the first parameter of the function.

backend/auth/firebase.go
```diff
@@ -15,36 +15,56 @@
...
 
+		if getDomain(userEmail) != *auth.AuthDomain {
+			http.Error(w, "You are not authorized to view this resource", http.StatusForbidden)
+			return
+		}
 		next(w, r)
 	}
 }
+
+func (auth FirebaseAuth) Authenticate(idToken string) (userEmail string, httpError *string) {
```
amdw, line 56: Why not just return an `error` rather than a `*string`? This is the conventional way to do error handling in Go.

src/app/services/notification.service.ts
```diff
@@ -23,6 +23,7 @@ import { Subject } from 'rxjs';
 export class NotificationService {
   public notification$: Subject<string> = new Subject<string>();
   public error$: Subject<string> = new Subject<string>();
+  public importantNotification$: Subject<string> = new Subject<string>();
```
amdw, line 26: Why don't we just use `error$` for this, and make these into modal dialogs?

As noted in #39, I'm already concerned that errors aren't as visible as they should be, and this would give us an opportunity to fix that as well.

src/app/app.component.ts
```diff
@@ -26,15 +27,28 @@ export class AppComponent {
...
+    });
+
+    this.notificationService.importantNotification$.subscribe(message => {
+      this.dialog.open(ModalComponent, {data: message});
     });
   }
   title = 'PeopleMath';
 }
 
+@Component({
```
amdw, line 47: Any reason we're not putting this in a separate file with its own template? That's what `ng generate component` will do by default.

backend/auth/firebase.go
```diff
@@ -15,36 +15,56 @@
...
+  "firebase.google.com/go/v4/auth"
+  "net/http"
+  "strings"
+  "time"
 )
 
 type FirebaseAuth struct {
 	FirebaseClient firebaseAuthClient
 	AuthTimeout    time.Duration
+	AuthDomain     *string
```
amdw, line 28: Why `*string` rather than just `string`?

backend/auth/auth.go
```diff
@@ -16,14 +16,25 @@ package auth
 
 import (
 	"net/http"
+	"strings"
 )
 
 type Auth interface {
-	Authenticate(next http.HandlerFunc) http.HandlerFunc
+	Authenticate(token string) (userEmail string, httpError *string)
+	Authorize(next http.HandlerFunc) http.HandlerFunc
```
amdw, line 24: This pattern is fine for now, but remember we're eventually going to want to make authorization decisions based on what the user is actually trying to do (e.g. read list of teams, write a period etc).

When we do that, we may actually want to make the authorization decisions inside the controller methods themselves (e.g. `handlePostPeriod`).

The pattern I'd expected we would follow for that is, for the HTTP endpoints requiring authentication and authorization, we'd use the authentication middleware to determine the authenticated `User`, and pass it as an additional parameter to the controller method. Then, the controller method would make a call to the `Auth` interface to make an authorization decision.

So I thought the `Auth` interface would gain a method for each permission query that's needed, e.g.:

```
CanReadTeamList(user User) bool
CanActOnTeam(user User, team Team, action Action) bool
```

where `action` is `read`, `write` or whatever. And these methods would then be called from the controller methods, passing in the authenticated user.

We don't really want to have to replicate the parsing of the URL path that we are already doing in the HTTP mux inside the `Authorize` method.

It's fine to go this way for now for this particular PR, but you may find yourself having to refactor this (including reversing some of the changes you made here) in your next PR.

backend/main_test.go
```diff
@@ -526,20 +531,39 @@ func (AuthClientStub) VerifyIDToken(ctx context.Context, idToken string) (*fireb
...
 	resp = makeHTTPRequest(req, handler, t)
-	checkResponseStatus(http.StatusOK, resp, t)
+	checkResponseStatus(http.StatusUnauthorized, resp, t)
+
+	req.Header.Set("Authorization", "Bearer reject")
+	resp = makeHTTPRequest(req, handler, t)
+	checkResponseStatus(http.StatusUnauthorized, resp, t)
+
+	// Testing with unauthorised user domain
+	authDomain = ""
```
amdw, line 556: The fact that this line of code has the side effect of changing the `AuthDomain` seen by `FirebaseAuth` is pretty confusing here!

We should definitely make the `FirebaseAuth` member a `string` (not a `*string`) and explicitly update the member here, i.e. `auth.AuthDomain = ""`.

amdw (changes requested):

backend/auth/firebase.go
Samrthi: I totally missed the fact that I could get the error message by calling err.Error() to put in the HTTP error. I updated it to use error as expected now.

src/app/services/notification.service.ts
Samrthi: done

src/app/app.component.ts
Samrthi: I initially didn't think this would work with a component at all and then didn't stop to reevaluate once I did use a component. I created a separate file now.

backend/auth/firebase.go
Samrthi: I struggled to get the tests to work because just doing `auth.AuthDomain = ""` doesn't seem to change the variable in the server (and/or handler?) so I thought using a pointer would make sense to allow the auth domain to be changed during runtime without the server having to be reinitialized. Seeing that that's neither something I have actually implemented nor something a user would likely want to do that was probably not a very useful consideration. I've changed it now.

backend/main_test.go
Samrthi: See previous. Let me know if there's any way to update the authDomain without reinitializing the server and handler that I'm missing.

backend/auth/firebase.go
Samrthi: done

backend/auth/firebase.go
Samrthi: I've updated it, let me know if I've done it correctly this time (still a bit confused by the context design pattern 😬).

backend/auth/auth.go
Samrthi: Okay, I'll take this into account for the next PR.

README.md
```diff
@@ -82,13 +82,19 @@ The primary supported way of running PeopleMath in a staging or production confi
...
 
-### Authentication
+### Authentication & Authorization
 
-To enable authentication using Firebase:
+A basic authorization mechanism has been implemented which specifies one email domain (like "google.com").
+Every user with an email address ending in that domain has read and write access.
+Any other user will not have access.
+
+To enable authorization with Firebase to authenticate users:
```
amdw, line 91: Nit: can you link to the Firebase Auth site here?

https://firebase.google.com/docs/auth

backend/main_test.go
```diff
@@ -526,20 +531,41 @@ func (AuthClientStub) VerifyIDToken(ctx context.Context, idToken string) (*fireb
...
 	resp = makeHTTPRequest(req, handler, t)
-	checkResponseStatus(http.StatusOK, resp, t)
+	checkResponseStatus(http.StatusUnauthorized, resp, t)
+
+	req.Header.Set("Authorization", "Bearer reject")
+	resp = makeHTTPRequest(req, handler, t)
+	checkResponseStatus(http.StatusUnauthorized, resp, t)
+
+	// Testing with unauthorised user domain
+	auth.AuthDomain = ""
```
amdw, line 556: Can you not just say `server.auth.AuthDomain = ""`? I don't understand why you'd have to create a new `Server` here.

amdw (changes requested):
Looks much better, thanks. Just a couple of small comments and I think we're good to go.

backend/auth/firebase.go
amdw: Looks good now. It's worth reading the following article if you haven't already: https://blog.golang.org/context

amdw (approved):

PR #61, opened 2020-09-03
User based auth

Samrthi:
The readme might need editing


backend/models/models.go
```diff
@@ -14,12 +14,42 @@
 
 package models
 
+type GeneralPermissions struct {
```
pallavshah, line 17: Can you add some inline comments to help understand what the struct fields are used for? For example, it's not very clear to me what's the purpose of AddTeamPermissions in TeamList.

backend/auth/firebase.go
```diff
@@ -16,55 +16,92 @@ package auth
...
+	for _, permission := range permissions {
+		if user.has(permission) {
+			isPermitted = true
+		}
 	}
+	return isPermitted
+}
+
+func (auth *FirebaseAuth) CanActOnTeam(user User, team models.Team, action string) bool {
+	permissions := getPermissionsFromTeam(team, action)
```
pallavshah, line 100: I would rename getPermissionsFromTeam() to getAllowedPrincipalsForTeam() and then call user.isPermitted(allowedPrincipals) which returns whether the user is permitted to perform the action based on the passed allowedPrincipals. Same for canActOnTeamList().

You are not doing anything wrong here. It's just that I feel we can make the code easier to understand with these changes. I would still defer to amedworth@ to make that call.

src/app/app.component.ts
```diff
@@ -35,7 +35,8 @@ export class AppComponent {
       this.snackBar.open(message, 'Dismiss');
     });
     this.notificationService.error$.subscribe(message => {
-      this.dialog.open(ModalComponent, {data: message});
+      const dismissInfo = ' Click anywhere to dismiss.';
```
pallavshah, line 38: nit: leading space in the message.

src/app/period/period.component.html
```diff
@@ -19,7 +19,7 @@ <h4 class="mat-h4" [class.warning]="committedAllocationsTooHigh()">{{committedAl
         <button mat-fab (click)="disableEditing()" color="primary"><mat-icon>done</mat-icon></button>
     </span>
     <span *ngIf="!isEditingEnabled">
-        <button mat-fab (click)="enableEditing()" color="primary"><mat-icon>edit</mat-icon></button>
+        <button mat-fab [disabled]="!userHasEditPermissions" (click)="enableEditing()" color="primary"><mat-icon>edit</mat-icon></button>
```
pallavshah, line 22: This is super nice and will really help the user experience. Thanks for adding this!

[No action needed]

src/app/teams/teams.component.ts
```diff
@@ -44,14 +47,20 @@ export class TeamsComponent implements OnInit {
...
-    ).subscribe((teams?: Team[]) => {
-      if (teams) {
-        this.teams = teams.map(t => new ImmutableTeam(t));
+    ).subscribe((teamList?: TeamList) => {
+      if (teamList?.teams) {
+        this.teams = teamList.teams.map(t => new ImmutableTeam(t));
       } else {
         this.teams = undefined;
       }
+      if (teamList?.addTeamPermissions) {
```
pallavshah, line 58: If I understood this correctly, the "AddTeamPermissions" is meant to act as an application level setting for whether the users are allowed to add team(s) or not. Is that true? If so, can you add appropriate documentation in the readme? If not, can you please explain what is the purpose of it?

backend/auth/auth.go
```diff
@@ -17,22 +17,42 @@ package auth
...
-	Authorize(next http.HandlerFunc) http.HandlerFunc
+	Authenticate(next http.HandlerFunc) http.HandlerFunc
+	CanActOnTeam(user User, team models.Team, action string) bool
+	CanActOnTeamList(user User, generalPermissions models.GeneralPermissions, action string) bool
+}
+
+const (
+	ActionRead         = "read"
+	ActionWrite        = "write"
+	ActionReadTeamList = "readTeamList"
```
amdw, line 33: These two actions seem redundant - can you just use `ActionRead` and `ActionWrite` in your `CanActOnTeamList` method?

backend/auth/firebase.go
```diff
@@ -16,55 +16,92 @@ package auth
...
 	AuthTimeout    time.Duration
-	AuthDomain     string
 }
 
 type firebaseAuthClient interface {
 	VerifyIDToken(ctx context.Context, idToken string) (*auth.Token, error)
 }
 
-func (auth *FirebaseAuth) Authorize(next http.HandlerFunc) http.HandlerFunc {
+func (auth FirebaseAuth) Authenticate(next http.HandlerFunc) http.HandlerFunc {
```
amdw, line 35: Shall we keep this as a pointer receiver?

backend/auth/firebase.go
```diff
@@ -16,55 +16,92 @@ package auth
...
-	ctx, cancel := context.WithTimeout(ctx, auth.AuthTimeout)
-	defer cancel()
+func getPermissionsFromTeam(team models.Team, action string) []models.Principal {
+	var permissions []models.Principal
+	if action == ActionRead {
+		permissions = team.Permissions.Read.Allow
+	} else if action == ActionWrite {
+		permissions = team.Permissions.Write.Allow
+	}
+	return permissions
```
amdw, line 71: Should do something if we get an unsupported action here. Candidates: `panic`, `log.Fatalf`, some other form of `log`.

backend/models/models.go
```diff
@@ -14,12 +14,42 @@
...
+type TeamPermissions struct {
+	Read  Permission `json:"read"`
+	Write Permission `json:"write"`
+}
+
+type Permission struct {
+	Allow []Principal `json:"allow"`
+}
+
+type Principal struct {
```
amdw, line 43: I know I suggested the name `Principal` myself, but I'm not quite sure it's strictly correct. I believe a principal is an entity that you can authenticate as, e.g. a user account or a service account, whereas this can cover things like email domains, which is a kind of predicate *through which* a given principal can get a given permission.

We should come up with an alternative name here: maybe `PrincipalMatcher`?

backend/auth/firebase.go
```diff
@@ -16,55 +16,92 @@ package auth
...
+
+		ctxWithUser := context.WithValue(r.Context(), "user", user)
+		next(w, r.WithContext(ctxWithUser))
 	}
 }
 
-func (auth *FirebaseAuth) Authenticate(ctx context.Context, idToken string) (userEmail string, err error) {
-	ctx, cancel := context.WithTimeout(ctx, auth.AuthTimeout)
-	defer cancel()
+func getPermissionsFromTeam(team models.Team, action string) []models.Principal {
```
amdw, line 64: We should come up with a better name for these. Maybe `getTeamAllowed` etc?

backend/auth/firebase.go
```diff
@@ -16,55 +16,92 @@ package auth
...
+	} else if action == ActionAddTeam {
+		permissions = generalPermissions.AddTeam.Allow
 	}
-	if !token.Claims["email_verified"].(bool) {
-		errorMessage = "User authentication failed, please verify your email address"
-		return "", errors.New(errorMessage)
+	return permissions
+}
+
+func (user User) has(permission models.Principal) bool {
```
amdw, line 84: We can probably move this logic to the `PrincipalMatcher`.

backend/auth/firebase.go
```diff
@@ -16,55 +16,92 @@ package auth
...
+	return (permission.Type == models.PrincipalTypeDomain && strings.ToLower(permission.ID) == strings.ToLower(user.domain)) ||
+		(permission.Type == models.PrincipalTypeEmail && strings.ToLower(permission.ID) == strings.ToLower(user.email))
+}
+
+func (user User) isPermitted(permissions []models.Principal) bool {
+	isPermitted := false
+	for _, permission := range permissions {
+		if user.has(permission) {
+			isPermitted = true
```
amdw, line 93: Can just short-cut `return true` here, then `return false` at the end if none of them matched.

backend/google_cds_store/google_cds_store.go
```diff
@@ -182,7 +182,7 @@ func (s *googleCDSStore) GetSettings(ctx context.Context) (models.Settings, erro
 	var result models.Settings
 	err := s.client.Get(ctx, key, &result)
 	if err == datastore.ErrNoSuchEntity {
-		result = models.Settings{}
+		result, err = s.CreateAndGetSettings(ctx, key)
```
amdw, line 185: We can probably just have an `UpsertSettings` method which updates settings in the database, creating a new settings object if not already present. That might be useful for other things later (e.g. updating the general settings).

If we're going to modify the behaviour of the `GetSettings` like this (which I think is a good idea) then we should probably do it in a separate PR.

backend/in_memory_storage/in_memory_storage.go
```diff
@@ -31,10 +31,46 @@ type inMemStore struct {
...
+func MakeInMemStore(defaultDomain string) storage.StorageService {
+	defaultPermissionList := models.Permission{Allow: []models.Principal{{
+		Type: models.PrincipalTypeDomain,
+		ID:   defaultDomain,
+	}}}
+	teamPermissions := models.TeamPermissions{
+		Read:  defaultPermissionList,
+		Write: defaultPermissionList,
+	}
+	userAEmail := "userA@userA.com"
```
amdw, line 43: My understanding is that these users are primarily for unit tests, while the passed-in `defaultDomain` is intended for people testing authentication manually when they run the backend in development using `--inmemstore` - is that right?

If so then I think we'd benefit from some comments to that effect.

backend/main.go
```diff
@@ -334,16 +414,19 @@ func (s *Server) handleImprove(w http.ResponseWriter, r *http.Request) {
 func main() {
 	var useInMemStore bool
 	var authMode string
-	var authDomain string
+	var defaultDomain string
 	flag.BoolVar(&useInMemStore, "inmemstore", false, "Use in-memory datastore")
 	flag.StringVar(&authMode, "authmode", "none", "Set authentication mode, either 'none' or 'firebase'")
-	flag.StringVar(&authDomain, "authdomain", "google.com", "The email domain with read and write access")
+	flag.StringVar(&defaultDomain, "defaultdomain", "none", "The domain that all team permissions are defaulted to")
```
amdw, line 420: If this only affects in-mem store, we should probably say so in the docstring.

backend/main_test.go
```diff
@@ -551,19 +553,196 @@ func TestFirebaseAuth(t *testing.T) {
 	req.Header.Set("Authorization", "Bearer reject")
 	resp = makeHTTPRequest(req, handler, t)
 	checkResponseStatus(http.StatusUnauthorized, resp, t)
+}
 
-	// Testing with unauthorised user domain
-	testAuth.AuthDomain = ""
+type UserAAuthClientStub struct {
```
amdw, line 558: You probably don't need two different types here - you can just have a single `struct` which holds the user you want to return.

amdw (changes requested):

README.md
```diff
@@ -84,17 +84,26 @@ The primary supported way of running PeopleMath in a staging or production confi
 
 ### Authentication & Authorization
 
-A basic authorization mechanism has been implemented which specifies one email domain (like "google.com").
-Every user with an email address ending in that domain has read and write access.
-Any other user will not have access.
+There are both general, app-wide permissions and team-specific permissions.
+The general permissions contain a list of users/domains that can read the list of all teams (`ReadTeamList`),
+and a list of users/domains that can add a new team (`AddTeam`).
+Each teams' permissions contain a list of users/domain that can read all the team's information (`Read`),
```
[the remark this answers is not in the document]
Samrthi: fixed

README.md
```diff
@@ -84,17 +84,26 @@ The primary supported way of running PeopleMath in a staging or production confi
...
+(the `Read` list will mirror the `ReadTeamList` list and the `Write` list will mirror the `AddTeam` list)
+
+To use authorization with [Firebase](https://firebase.google.com/docs/auth) to authenticate users:
 
 * Set `requireAuth` to `true` in `src/environments/environment.prod.ts`
 * Run backend with authmode flag `--authmode firebase`
 * Download your [Firebase config file](https://support.google.com/firebase/answer/7015592#web) and copy the text into the `firebaseConfig.ts` file in the same folder.
 * Download your [Firebase credentials](https://firebase.google.com/docs/admin/setup#initialize-sdk) and export them via `export GOOGLE_APPLICATION_CREDENTIALS="/path/to/your/credentials.json"`
-* Set the email domain for the authorization using the `authdomain` flag. The default is `google.com`
+* Set the general permissions in Pantheon after starting the AppEngine instance
```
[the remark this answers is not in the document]
Samrthi: 👍🏻

src/app/app.component.ts
Samrthi: The leading space is because the message it is appended to will end with a "."
Let me know if there is a better/more readable way to do this.

backend/main_test.go
Samrthi: Yep, should have caught that.

backend/main.go
Samrthi: 👍🏻

backend/in_memory_storage/in_memory_storage.go
Samrthi: I've added a comment about the users, let me know if that is what you were thinking.

backend/auth/firebase.go
Samrthi: Ah yes, this was a relic from when this chunk was still part of a larger function but I should have caught this.

backend/auth/auth.go
Samrthi: Yep, makes more sense.

backend/models/models.go
Samrthi: I thought about this for a good while and have decided on calling it UserMatcher because really user is the only Principal we are using are users so I feel like this makes sense. Let me know what you think.

backend/auth/firebase.go
Samrthi: Done, I did have to move a few things around to avoid a cyclical import though so there is now methods in models.go, let me know if that is okay.

backend/auth/firebase.go
Samrthi: Yes, I don't see any reason against making this change as a precaution.

backend/auth/firebase.go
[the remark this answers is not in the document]
Samrthi: 👍🏻

backend/auth/firebase.go
Samrthi: I find getTeamAllowed a bit confusing because it's not clear what is being returned, I've changed it to getTeamAllowedUsers now which would obviously really be get team allowed user matchers but I did want to keep the function name shorter, let me know what you think.

backend/auth/firebase.go
Samrthi: Good point! I've changed it to getTeamAllowedUsers now with users essentially being short for user matcher, the new name I have to the principal. I will check with amedworth@ if this name is okay.

backend/main.go
```diff
@@ -204,19 +244,32 @@ func (s *Server) handleGetAllPeriods(w http.ResponseWriter, r *http.Request) {
 	ctx, cancel := context.WithTimeout(r.Context(), s.storeTimeout)
 	defer cancel()
 
-	periods, found, err := s.store.GetAllPeriods(ctx, teamID)
+	team, found, err := s.store.GetTeam(ctx, teamID)
 	if err != nil {
-		log.Printf("Could not retrieve periods for team '%s': error: %s", teamID, err)
-		http.Error(w, fmt.Sprintf("Could not retrieve periods for team '%s' (see server log)", teamID), http.StatusInternalServerError)
+		http.Error(w, "Authorization failed due to internal server error", http.StatusInternalServerError)
```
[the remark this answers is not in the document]
Samrthi: 👍🏻

backend/models/models.go
Samrthi: Let me know if the comments make it a bit clearer now.

src/app/teams/teams.component.ts
Samrthi: I have tried to explain it in the README, could you let me how I could improve what I've written to make this clearer?

backend/auth/firebase.go
Samrthi: 👍🏻

backend/google_cds_store/google_cds_store.go
Samrthi: Okay, I've reset the file now.

backend/main.go
```diff
@@ -146,44 +159,71 @@ func (s *Server) handleGetTeam(w http.ResponseWriter, r *http.Request) {
...
+	if s.auth.CanActOnTeamList(r.Context().Value("user").(auth.User), permissions, auth.ActionAddTeam) {
+		if reflect.DeepEqual(team.Permissions, models.TeamPermissions{}) {
+			team.Permissions.Read = permissions.ReadTeamList
+			team.Permissions.Write = permissions.AddTeam
+		}
+
+		err := s.store.CreateTeam(ctx, team)
+		if err != nil {
+			log.Printf("Could not create team: error: %s", err)
+			http.Error(w, "Could not create team (see server log)", http.StatusInternalServerError)
```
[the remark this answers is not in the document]
Samrthi: I have left it for now because this was part of Andrew's implementation but it would probably make sense to implement a more useful log application wide in the future.

src/app/team.ts
```diff
@@ -15,23 +15,59 @@
...
+}
+
+export class Permission {
+  constructor(
+    public allow: Principal[]
+  ) {
+  }
+}
+
+export class Principal {
```
amdw, line 47: Can we rename this to be consistent with what we did in the backend?

src/app/services/notification.service.ts
```diff
@@ -23,6 +23,7 @@ import { Subject } from 'rxjs';
 export class NotificationService {
   public notification$: Subject<string> = new Subject<string>();
   public error$: Subject<string> = new Subject<string>();
+  public importantNotification$: Subject<string> = new Subject<string>();
```
amdw, line 26: This seems to be unused?

src/app/period/period.component.ts
```diff
@@ -270,6 +275,22 @@ export class PeriodComponent implements OnInit {
...
+          const principalTypeEmail = 'email';
+          const principalTypeDomain = 'domain';
+          team.teamPermissions.write.allow.forEach(permission => {
+            if ((permission.type === principalTypeDomain && permission.id.toLowerCase() === userDomain?.toLowerCase()) ||
+              (permission.type === principalTypeEmail && permission.id.toLowerCase() === userEmail?.toLowerCase())) {
+              this.userHasEditPermissions = true;
+            }
+          });
+          if (!this.userHasEditPermissions) {
+            this.notificationService.error$.next('You do not have editing priviliges.');
```
amdw, line 291: Why do we show an error when a user has read-only access? Opening a read-only team is perfectly normal - it's not an error.

Can't we just leave the edit button greyed out, maybe with a mouseover that says the user doesn't have permission?

src/app/period/period.component.ts
```diff
@@ -44,16 +46,19 @@ export class PeriodComponent implements OnInit {
...
   readonly AggregateBy = AggregateBy;
- 
+
   constructor(
     private storage: StorageService,
     private route: ActivatedRoute,
     private dialog: MatDialog,
     private snackBar: MatSnackBar,
     private changeDet: ChangeDetectorRef,
+    public authService: AuthService,
```
amdw, line 60: Can this be private?

src/app/app.component.ts
amdw: Can we just have a "close" button, rather than editing the message?

backend/models/models.go
amdw: The comments definitely help, though I think `AddTeamPermissions` would be better named something like `HasAddTeamPermissions` or `CanAddTeams`.

Normally boolean-typed things are best given names starting with a predicate verb like "is", "can", or "has".

backend/main.go
```diff
@@ -334,16 +414,19 @@ func (s *Server) handleImprove(w http.ResponseWriter, r *http.Request) {
...
+	flag.StringVar(&defaultDomain, "defaultdomain", "none", "When using inmemstore: the domain that all team permissions are defaulted to")
 	flag.StringVar(&authMode, "authmode", "none", "Set authentication mode, either 'none' or 'firebase'")
-	flag.StringVar(&authDomain, "authdomain", "google.com", "The email domain with read and write access")
 	flag.Parse()
 
 	var store storage.StorageService
 	if useInMemStore {
 		log.Printf("Using in-memory store per command-line flag")
-		store = in_memory_storage.MakeInMemStore()
+		if defaultDomain == "none" {
```
amdw, line 426: Why not just set the default to `google.com`?

backend/main.go
```diff
@@ -280,23 +353,30 @@ func (s *Server) handlePutPeriod(w http.ResponseWriter, r *http.Request) {
...
 	period.LastUpdateUUID = uuid.New().String()
 	ctx, cancel := context.WithTimeout(r.Context(), s.storeTimeout)
 	defer cancel()
-	err := s.store.UpdatePeriod(ctx, teamID, period)
-	if err != nil {
-		log.Printf("Could not update period '%s' for team '%s': error: %s", periodID, teamID, err)
-		http.Error(w, fmt.Sprintf("Could not update period '%s' for team '%s' (see server log)", periodID, teamID), http.StatusInternalServerError)
-		return
+
+	if s.auth.CanActOnTeam(r.Context().Value("user").(models.User), team, auth.ActionWrite) {
```
amdw, line 369: Can we extract the user to a variable here, just to simplify this line slightly?

Same comment in a few other places.

backend/main.go
```diff
@@ -280,23 +353,30 @@ func (s *Server) handlePutPeriod(w http.ResponseWriter, r *http.Request) {
 	if !ok {
 		return
 	}
-	if !s.ensureTeamExistence(w, r, teamID, true) {
+	team, exists := s.ensureTeamExistence(w, r, teamID, true)
+	if !exists {
 		return
 	}
+	log.Print(team)
```
amdw, line 360: Why are we logging this?

backend/main.go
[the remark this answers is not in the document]
amdw: Yes, I filed #66 for this. This is fine for now.

backend/in_memory_storage/in_memory_storage.go
```diff
@@ -31,10 +31,48 @@ type inMemStore struct {
...
+	defaultPermissionList := models.Permission{Allow: []models.UserMatcher{{
+		Type: models.UserMatcherTypeDomain,
+		ID:   defaultDomain,
+	}}}
+	teamPermissions := models.TeamPermissions{
+		Read:  defaultPermissionList,
+		Write: defaultPermissionList,
+	}
+
+	// Sample users that are used in unit tests
```
amdw, line 44: Ideally these users shouldn't be present when people are using the in-mem store for local manual testing and development. We should probably have another constructor, called by the unit tests, that adds them.

It's OK to go this way for now, but can you at least add a TODO to the code for this?

backend/in_memory_storage/in_memory_storage.go
```diff
@@ -31,10 +31,48 @@ type inMemStore struct {
 	settings models.Settings
 }
 
-func MakeInMemStore() storage.StorageService {
+func MakeInMemStore(defaultDomain string) storage.StorageService {
```
amdw, line 34: Can you add a brief comment to this function signature to explain what `defaultDomain` is for?

README.md
```diff
@@ -84,17 +84,82 @@ The primary supported way of running PeopleMath in a staging or production confi
...
+  }
+}
+```
 
 Currently, the only authentication method enabled on Firebase is Google Sign-in.
 
+To use in memory storage with authentication and authorization,
+set the defaultdomain flag with your account's domain `--defaultdomain yourdomain.com`
+
+Because the permissions have to be set manually, there is currently no way to use authentication and authorization with the Google Cloud emulator.
```
amdw, line 165: I'm not sure this is true, is it? Why can't you just connect to Firebase Auth from your development machine using service account credentials, and manually insert permissions into the (Cloud Datastore emulator based) database like you have suggested doing above?

OK, you won't be able to use the Pantheon UI, but you should be able to post these JSON entries into the API using `curl` or equivalent.

README.md
```diff
@@ -84,17 +84,82 @@ The primary supported way of running PeopleMath in a staging or production confi
 
 ### Authentication & Authorization
```
amdw, line 85: We need to reorganise this a bit, as this content doesn't all belong in the App Engine section. It should be possible for people to use Firebase Auth using the code you've built, without running on App Engine.

Don't worry about this for now though - I'll work on this later.

amdw (changes requested):

backend/models/models.go
amdw: SGTM, thank you.

src/app/teams/teams.component.ts
amdw: As noted above, I think renaming `addTeamPermissions` would really help here.

backend/auth/firebase.go
amdw: Thanks, this is fine.

backend/in_memory_storage/in_memory_storage.go
amdw: Yes, this is better, thanks - though see comments elsewhere.

backend/models/models.go
Samrthi: 👍🏻

src/app/app.component.ts
Samrthi: Yes, I didn't realise just how easy that is to implement with angular.

src/app/teams/teams.component.ts
Samrthi: 👍🏻

src/app/team.ts
Samrthi: 👍🏻

src/app/services/notification.service.ts
Samrthi: It is, I forgot to update after changing the app.component.ts. Fixed now.

src/app/period/period.component.ts
Samrthi: Whoops, I had this as a notification and then didn't think further when I merged important_notification and error.
I looked into making a tooltip, but creating a tooltip for a disabled element is not really supported in angular. I've left it for now.

src/app/period/period.component.ts
Samrthi: Yes.

backend/main.go
Samrthi: I'm sure I had a reason why I wrote these lines in the first place but I can't see them being useful at all now, so I've changed the default to google.com now

backend/main.go
Samrthi: 👍🏻

backend/main.go
Samrthi: 😬 oops, this was definitely for debugging.

README.md
Samrthi: Ah, okay, my bad! I was trying to find answers online whether you can update the database when using the emulator and couldn't find any answers but what you're saying is probably true. I now found some posts that indicate it is possible, but I don't understand it well enough to be able to update the README.

backend/in_memory_storage/in_memory_storage.go
Samrthi: I added a parameter in the constructor determining whether the test users should be added. It's probably not ideal but should be easy to refactor because the code for the test users is all separate now.

backend/in_memory_storage/in_memory_storage.go
Samrthi: 👍🏻

Issue #65, opened 2020-09-11
Redesign assignment dialog


amdw:
The design of the dialog to assign people to objectives needs a bit of a rethink.

A few ideas:

* The "all" and "remaining" buttons are a bit confusing, and should probably be merged into a single button, which reassigns all the unassigned work to a given person (up to the maximum they have available).
* Buttons should be consistently greyed out when pressing them makes no sense. For example, it shouldn't be possible to press "+1" when an objective is fully assigned.
* We should consider not showing people who don't have any resources available, as it's not possible to usefully interact with these, and they just clutter the view.
* We should consider a way to get to the assignment dialog from the edit-objective dialog. For example currently, when an objective has an estimate of zero and you want to change that and then assign it, you have to click to edit the objective, change the estimate, then close the dialog (and save) and then press assign.

DakshSaini:
Hi @amdw, can I take this issue?

amdw:
Hi @DakshSaini - thank you very much for offering! I have no immediate plans to work on this myself, so would be very happy to take contributions on it.

The one thing I would say is, it would really help if we could break this down into as many small pull requests as reasonably possible, starting with the parts that are easiest to implement and agree upon.

For example, I think each of the bullet points in the original comment should be a separate pull request. As far as I can see, each of these changes would have value on its own (I'm by no means saying you'd have to implement them all!).

Happy to hear any thoughts you may have on the best order to go about things.

Feel free to send me some PRs. I'll review them as quickly as I reasonably can. Thanks a lot!

DakshSaini:
Sounds great, @amdw. Thanks!

> For example, I think each of the bullet points in the original comment should be a separate pull request. As far as I can see, each of these changes would have value on its own (I'm by no means saying you'd have to implement them all!).

I agree!
> Happy to hear any thoughts you may have on the best order to go about things.
> 

I think sequential order, starting with first bullet point should be fine.

amdw:
I think I'll stop here.

Items 1-3 on the original list are done. Item 4 would be nice to have but I'm not convinced it's that inconvenient to save the objective and then click "assign", versus the inconvenience of implementing a button that takes you straight from one to the other.

PR #90, opened 2020-10-06
Redesign assignment dialog - remove 'all' button

DakshSaini:
This pull request addresses bullet point 1 in issue #65

### DIFF
--- a/src/app/assignment-dialog/assignment-dialog.component.html
+++ b/src/app/assignment-dialog/assignment-dialog.component.html
@@ -19,7 +19,6 @@
             <td mat-cell *matCellDef="let element">
                 <button mat-button (click)="assignNone(element)" color="primary">None</button>
                 <button mat-button (click)="assignAll(element)" color="primary">All</button>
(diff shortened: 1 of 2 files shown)


DakshSaini:
@amdw, can I get a review? Thanks!

amdw (changes requested):
Many thanks for sending this pull request!

In terms of *functionality*, I think it's actually the "remaining" button we want to keep.

This will assign all the unassigned work on the objective to the selected person, up to the maximum they have available. If the objective is completely unassigned, then it will have the same effect as the "all" button used to have.

I'm open to the possibility of renaming the button, but I think "remaining" is good enough for now - what do you think?

If you agree, then let's just remove the "all" button instead of the "remaining" button?

Thanks, again, for contributing to PeopleMath!

DakshSaini:
That makes sense! I Fixed it, Thank you @amdw

amdw:
Thank you very much @DakshSaini!

PR #91, opened 2020-10-07
Redesign assignment dialog - do not show people with no resources

DakshSaini:
Not showing people who don't have any resources available in the assignment dialog
Issue #65

### DIFF
--- a/src/app/objective.ts
+++ b/src/app/objective.ts
@@ -88,7 +88,7 @@ export class ImmutableObjective {
   readonly groups: readonly ImmutableObjectiveGroup[];
   readonly tags: readonly ImmutableObjectiveTag[];
   readonly assignments: readonly ImmutableAssignment[];
-  
+
   private constructor(o: ImmutableObjectiveIF) {
     this.name = o.name;
     this.resourceEstimate = o.resourceEstimate;
@@ -97,6 +97,7 @@ export class ImmutableObjective {
     this.groups = o.groups;
     this.tags = o.tags;
     this.assignments = o.assignments;
+    this.assignments.length > 0;
   }
 
   static fromObjective(objective: Objective): ImmutableObjective {
--- a/src/app/objective.ts
+++ b/src/app/objective.ts
@@ -28,7 +28,7 @@ export class ImmutableObjectiveGroup {
 
   get groupType(): string { return this._groupType; }
   get groupName(): string { return this._groupName; }
-  
+
   constructor(g: ObjectiveGroup) {
     this._groupType = g.groupType;
     this._groupName = g.groupName;
@@ -97,7 +97,6 @@ export class ImmutableObjective {
     this.groups = o.groups;
     this.tags = o.tags;
     this.assignments = o.assignments;
-    this.assignments.length > 0;
   }
 
   static fromObjective(objective: Objective): ImmutableObjective {
--- a/src/app/objective/objective.component.ts
+++ b/src/app/objective/objective.component.ts
@@ -37,10 +37,24 @@ export class ObjectiveComponent implements OnInit {
   @Output() onMoveBucket = new EventEmitter<[ImmutableObjective, ImmutableObjective, ImmutableBucket]>();
   @Output() onDelete = new EventEmitter<ImmutableObjective>();
   @Output() onChanged = new EventEmitter<[ImmutableObjective, ImmutableObjective]>();
+  assignmentData: PersonAssignmentData[] = [];
 
   constructor(public dialog: MatDialog) { }
 
   ngOnInit() {
+    this.createAvailableAssignmentData();
+  }
+
+  createAvailableAssignmentData(): void {
+    this.unallocatedTime!.forEach((unallocated, personId) => {
+      let currentAssignment = this.currentAssignment(personId);
+      const availableWeeks = unallocated + currentAssignment;
+      if (availableWeeks > 0) {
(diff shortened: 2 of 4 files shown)


DakshSaini:
@amdw, can I get a review? Thanks!

amdw (changes requested):
This looks exactly right, in terms of excluding people with no available time from the assignment dialog.

However, we now need to consider the case where there are actually no people with any available time, and so the `assignmentData` array ends up empty. With this change, what happens in that case is that the assignment dialog pops up with an empty table.

I think it would be better to grey out the "assign" buttons when there is nobody with available time. Is that something you feel able to do as part of this PR?

Many thanks for contributing to PeopleMath!

amdw (changes requested):
I'd prefer not to cache the `PeopleAssignmentData[]` as a property of the component like this, because I don't think it's guaranteed that `ngOnInit()` will be called if the `@Input`s of the component are assigned new values.

I think Angular is allowed to reuse the old instance of the component, and I believe `ngOnInit()` is only called once in the lifetime of the component.

(Even if `createAvailableAssignmentData` were guaranteed to be called whenever the `@Input`s change, this code still wouldn't do the right thing, because it just pushes to `this.assignmentData` array without re-initializing it to empty beforehand. So calling `createAvailableAssignmentData` multiple times will cause more and more data to build up in `assignmentData`.)

I think we should add a method `hasPeopleAvailable()` which returns a `boolean` indicating if anyone is available to work on this objective. This should return `true` if there is already at least one non-zero resource assignment to the objective, *or* there is at least one person with unallocated resources in `unallocatedTime`.

Then, we can simply modify `enableAssignButton()` to require that `hasPeopleAvailable()` returns `true`.

Also, rather than the change you made to the HTML file, let's please modify the first line of the `assign()` function so that it returns immediately if `enableAssignButton()` is `false` (as opposed to `this.isEditingEnabled` as now).

Many thanks for contributing!

src/app/objective/objective.component.ts
```diff
@@ -37,12 +37,24 @@ export class ObjectiveComponent implements OnInit {
...
   @Output() onChanged = new EventEmitter<[ImmutableObjective, ImmutableObjective]>();
-  
+
   constructor(public dialog: MatDialog) { }
 
   ngOnInit() {
   }
 
+  hasPeopleAvailable() {
+    if (this.objective!.assignments.length) {
```
amdw, line 47: Strictly speaking, we should probably check that the assignments have positive value, i.e. something like:

```typescript
if (this.objective!.assignments.find(a => a.commitment > 0) {
    return true;
}
```

src/app/objective/objective.component.ts
```diff
@@ -37,12 +37,24 @@ export class ObjectiveComponent implements OnInit {
...
+      return true;
+    }
+    for (const weeks of this.unallocatedTime!.keys()) {
```
amdw, line 50: I think this could be done more simply as follows:

```typescript
if (Array.from(this.unallocatedTime!.values()).find(t => t > 0)) {
    return true;
}
```

In fact this whole function could probably be replaced by a single line:

```typescript
return this.objective!.assignments.find(a => a.commitment > 0) || Array.from(this.unallocatedTime!.values()).find(t => t > 0);
```

amdw (changes requested):
I'm really sorry for the slow response on this. Many thanks for these latest changes!

src/app/objective/objective.component.ts
```diff
@@ -121,6 +137,6 @@ export class ObjectiveComponent implements OnInit {
   }
 
   enableAssignButton(): boolean {
-    return !!this.isEditingEnabled && this.objective!.resourceEstimate > 0;
+    return this.hasPeopleAvailable() && !!this.isEditingEnabled && this.objective!.resourceEstimate > 0;
```
amdw, line 140: Minor performance consideration: we should probably check `hasPeopleAvailable()` last, since the other checks are extremely cheap and may avoid the need to call it:

```typescript
return !!this.isEditingEnabled && this.objective!.resourceEstimate > 0 && this.hasPeopleAvailable();
```

src/app/objective/objective.component.ts
DakshSaini: Fixed it, thanks!

src/app/objective/objective.component.ts
DakshSaini: Fixed it, thanks!

src/app/objective/objective.component.ts
DakshSaini: Fixed it, thanks!

src/app/objective.ts
```diff
@@ -97,6 +97,7 @@ export class ImmutableObjective {
     this.groups = o.groups;
     this.tags = o.tags;
     this.assignments = o.assignments;
+    this.assignments.length > 0;
```
amdw, line 100: This line doesn't do anything. Was it left here by mistake?

amdw (changes requested):
Thanks! This looks much better. If you can just remove the one line that doesn't do anything, I think we can merge this.

src/app/objective.ts
DakshSaini: Sorry, fixed it now.

amdw (approved):
Looks good, thanks! I'll merge it tomorrow after a quick test.

PR #92, opened 2020-11-08
Add new colour for zero resource estimate

DakshSaini:
Hi @amdw, this Pull Request addresses Issue #3 

Screenshot:
Screen Shot


amdw (changes requested):
Thanks for sending this!

I'd prefer not to depend on the fact that the new CSS class is defined later in the `.css` file.

Can we please avoid applying both `.unassigned` and `.zero-resource-estimate` to the same element at the same time? Thanks!

Issue #58, opened 2020-08-25
Add the ability to delete a bucket


amdw:
This is quite a destructive operation, so careful verification would be needed before actually doing it (especially if there are actual objectives in the bucket). But it's still something that needs to be done sometimes.

amdw:
Fixed thanks to @DakshSaini !

PR #93, opened 2020-11-09
Add ability to delete a bucket

DakshSaini:
Hi @amdw, this closes issue #58.
Thanks!

### DIFF
--- a/src/app/period.ts
+++ b/src/app/period.ts
@@ -119,6 +119,10 @@ export class ImmutablePeriod implements ImmutablePeriodIF {
     return new ImmutablePeriod({...this, lastUpdateUUID: lastUpdateUUID});
   }
 
+  withNewBuckets(newBuckets: readonly ImmutableBucket[]): ImmutablePeriod {
+    return new ImmutablePeriod({...this, buckets: newBuckets});
+  }
+
   withNewBucket(bucket: ImmutableBucket): ImmutablePeriod {
     return new ImmutablePeriod({...this, buckets: this.buckets.concat([bucket])});
   }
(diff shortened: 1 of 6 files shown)


src/app/period.ts
```diff
@@ -119,6 +119,10 @@ export class ImmutablePeriod implements ImmutablePeriodIF {
     return new ImmutablePeriod({...this, lastUpdateUUID: lastUpdateUUID});
   }
 
+  withNewBuckets(newBuckets: readonly ImmutableBucket[]): ImmutablePeriod {
```
amdw, line 122: I believe this function can be private - is that right?

I also think if we're going to abstract this code, there are a number of other places in the file we can use it.

src/app/period.ts
```diff
@@ -153,6 +157,20 @@ export class ImmutablePeriod implements ImmutablePeriodIF {
     return new ImmutablePeriod({...this, buckets: newBuckets});
   }
 
+  private bucketIndex(bucket: ImmutableBucket): number {
+    return this.buckets.findIndex(b => b === bucket);
+  }
+
+  withBucketDeleted(bucket: ImmutableBucket): ImmutablePeriod {
```
amdw, line 164: We need to add tests for this function in period.spec.ts.

amdw (changes requested):
Thanks for sending this! I applied a few changes on top (added tests, minor refactoring) and created the new pull request #97 - please could you comment `@googlebot I consent.` on that pull request, to indicate your consent to your commit being merged there?

I'll merge that, and then we can close this one.

Thanks again!

amdw:
Actually, to save time, I'll merge this now and then add my commits on top.

PR #97, opened 2020-12-01
Add ability to delete a bucket (closes #58)

### LINKED ISSUE #58: Add the ability to delete a bucket

amdw:


amdw:
Obsoleted by #98.

PR #110, opened 2021-01-13
Issue 17

r-sreesaran:
This is a partial fix for the issue. 
1) The main problem with this fix is whenever the objectives are re arranged the entire view refreshes  and the cumulative sum is also calculated.

Any idea on how to resolve this or this is there any better way to achieve this ?.

And also instead of the background being coloured, I am display cumulative sum with coloured font. I though it would be more mellow.

Thanks
R.Sreesaran


src/app/bucket/bucket.component.css
```diff
@@ -4,3 +4,4 @@
 .reorder-icon { vertical-align: middle; padding-right: 10px; padding-bottom: 15px; }
 .app-mat-list .mat-list-item { height: auto; min-height: 40px; }
 .mat-list-item app-objective { padding-bottom: 15px; }
+.width { width: 100%;}
```
amdw, line 7: Let's not use such a generic class name as `width`, please. Maybe you can just add this `width: 100%` to the `.mat-list-item app-objective` line above?

src/app/bucket/bucket.component.html
```diff
@@ -24,27 +24,33 @@
     <mat-list-item *ngFor="let objective of bucket!.objectives">
     <app-objective
         [objective]="objective" [unit]="unit"
+        [sum]="cumilativeSum"
```
amdw, line 27: spelling: `cumulative` (also needs fixing elsewhere)

src/app/bucket/bucket.component.html
```diff
@@ -24,27 +24,33 @@
...
-        (onChanged)="onObjectiveChanged($event[0], $event[1])"></app-objective>
+        (onChanged)="onObjectiveChanged($event[0], $event[1])" class="width"></app-objective>
     </mat-list-item>
 </mat-list>
 <mat-list dense *ngIf="showOrderButtons" cdkDropList (cdkDropListDropped)="reorderDrop($event)" class="app-mat-list">
     <mat-list-item *ngFor="let objective of bucket!.objectives" cdkDrag>
         <i class="material-icons reorder-icon" cdkDragHandle>reorder</i>
         <app-objective
             [objective]="objective" [unit]="unit"
+            (cumiliativeSum)=setCumilativeSum($event)
```
amdw, line 44: As noted elsewhere, we shouldn't call methods with side effects in the template.

src/app/bucket/bucket.component.ts
```diff
@@ -43,6 +43,7 @@ export class BucketComponent implements OnInit {
   @Output() onMoveObjectiveBucket = new EventEmitter<[ImmutableObjective, ImmutableBucket, ImmutableObjective, ImmutableBucket]>();
   @Output() onChanged = new EventEmitter<[ImmutableBucket, ImmutableBucket]>();
   @Output() onDelete = new EventEmitter<ImmutableBucket>();
+  @Input() cumilativeSum?:number=0;
```
amdw, line 46: Please add this at the end of the list of `@Input()`s, so that all the inputs are together and all the outputs are together.

Also: do we need an explicit default here? I'd expect this to be defined on every objective because of the way they are constructed.

src/app/bucket/bucket.component.ts
```diff
@@ -161,4 +162,10 @@ export class BucketComponent implements OnInit {
     let total = this.resourcesAllocated();
     return total ? this.committedResourcesAllocated() / total : 0;
   }
+
+  setCumilativeSum(sum:number) {
```
amdw, line 166: I don't think we need this set method.

src/app/objective/objective.component.css
```diff
@@ -9,3 +9,10 @@
 .assign-button-enabled a { color: #674ab7; text-decoration: none; }
 .assign-button-enabled a:hover { text-decoration: underline; }
 .assign-button-disabled { color: grey; }
+.sum-all { display: flex; justify-content: center;   width: 10%; align-items: center;}
```
amdw, line 12: I could be wrong, but I think we can merge the `sum-all` and `sum` classes. We should probably call it something more descriptive like `.resource-csum`.

src/app/objective/objective.component.css
```diff
@@ -9,3 +9,10 @@
 .assign-button-enabled a { color: #674ab7; text-decoration: none; }
 .assign-button-enabled a:hover { text-decoration: underline; }
 .assign-button-disabled { color: grey; }
+.sum-all { display: flex; justify-content: center;   width: 10%; align-items: center;} 
+.sum { font-weight: bold;}
+.objective {float: left; width: 90%;}
+.objective-all {display: flex;}
+.background-safe { color: grey; }
```
amdw, line 16: This class doesn't affect the background, so shouldn't have `background` in the name.

The name should reflect the actual meaning we're going for; I'd suggest something like `resource-csum-ok`, `resource-csum-marginal`, and `resource-csum-excess` (open to better suggestions).

src/app/objective/objective.component.ts
```diff
@@ -147,4 +152,23 @@ export class ObjectiveComponent implements OnInit {
...
   }
+  displaySum() {
+    console.log("The sum is ");
+    let sum = Number(this.sum!) + Number(this.objective!.resourceEstimate)
+    console.log(sum);
+    this.cumiliativeSum.emit(sum);
+    return  sum;
+  }
+
+  setBackground() {
```
amdw, line 163: This isn't a set method. It should be called something like `getCumulativeSumClass()`.

It should also have a return type annotation.

src/app/objective/objective.component.ts
```diff
@@ -37,10 +37,15 @@ export class ObjectiveComponent implements OnInit {
   @Output() onMoveBucket = new EventEmitter<[ImmutableObjective, ImmutableObjective, ImmutableBucket]>();
   @Output() onDelete = new EventEmitter<ImmutableObjective>();
   @Output() onChanged = new EventEmitter<[ImmutableObjective, ImmutableObjective]>();
+  @Input() sum?:number;
+  @Input() bucketAllocationLimit?:number;
+  @Output() cumiliativeSum= new EventEmitter<number>();
```
amdw, line 42: I don't think we need these two things. We just need the cumulative sum (to display) and the bucket limit (to decide the colour).

src/app/objective/objective.component.ts
```diff
@@ -147,4 +152,23 @@ export class ObjectiveComponent implements OnInit {
...
+  displaySum() {
+    console.log("The sum is ");
+    let sum = Number(this.sum!) + Number(this.objective!.resourceEstimate)
+    console.log(sum);
+    this.cumiliativeSum.emit(sum);
+    return  sum;
+  }
+
+  setBackground() {
+    if(this.totalSum<this.bucketAllocationLimit!) {
```
amdw, line 164: I think the way to calculate this would be:

```typescript
if (this.cumulativeSum <= this.bucketAllocationLimit) {
    return "resource-csum-ok";
}
else if (this.cumulativeSum - this.objective.resourceEstimate <= this.bucketAllocationLimit) {
    return "resource-csum-marginal";
}
else {
    return "resource-csum-excess";
}
```

I don't believe we should need any `Number()` conversions. These values should always be defined, so you can just use `!` as necessary if you need to make the Typescript type-checker happy.

src/app/objective/objective.component.ts
```diff
@@ -147,4 +152,23 @@ export class ObjectiveComponent implements OnInit {
         (this.objective!.resourceEstimate > 0 || this.objective!.assignments.length > 0) &&
         this.hasPeopleAvailable();
   }
+  displaySum() {
```
amdw, line 155: I don't believe this function should be needed.

amdw (changes requested):
Thanks very much for having a go at this!

I agree with you - this doesn't behave correctly when the objectives change and the cumulative sums need to be recalculated.

I think the core of the problem is that you're trying to do an operation with side effects in the middle of your template. This is going to cause many difficulties. Templates shouldn't call functions with side effects, they should just read data out of the local component.

One suggestion is:

* Define a new interface inside `bucket.component.ts` called something like `DisplayedObjective`, with fields `objective: ImmutableObjective` and `cumulativeSum: number`.
* Define a new method on `BucketComponent` called something like `displayedObjectives(): DisplayedObjective[]` that generates the `DisplayedObjective` list and calculates the cumulative sums.
* In the `bucket.component.html` template, iterate over `displayedObjectives()` rather than the `bucket!.objectives` as now. Pass the `objective` and `cumulativeSum` into the `app-objective` component.
* Then you just need a two new `@Input()`s called `cumulativeSum` and `bucketAllocationLimit` on the `ObjectiveComponent`, which you can use to display the cumulative sum as you need to.

See also a few comments inline.

r-sreesaran:
I have made all the changes adapted. Please do check and let me know if anything else needs to be done.
Thanks for your time in reviewing the code.

r-sreesaran:
just now I noticed, there is am issue. The dragging and dropping seems to broken.
i will try to resolve that. If you have any clues on how to resolve please do suggest.

amdw:
To fix the drag-and-drop, probably you need to add `[cdkDragData]="displayObjective.objective"` to the `mat-list-item` element, alongside `cdkDrag`. Not sure, give it a try.

amdw:
I tried a few experiments, and for the life of me, I could not get drag-and-drop to work when the `mat-list-item`s are generated by an `*ngFor` over the result of a function, like `displayObjectives()`.

The following works for me:

* Add a member of type `DisplayObjective[]` to the `BucketComponent` - something like `objectives: DisplayObjective[] = []`
* In `ngOnInit()`, initialize this by `this.objectives = this.displayObjectives()`
* In the template, do `*ngFor="let displayObjective of objectives"`

This isn't a great solution, though. For correctness, this should probably be done in `ngOnChanges()`, which might cause performance issues in the future. Regardless, I don't see why it should be necessary to do that in order to get drag-and-drop to work. To me this feels like a CDK bug; at the very least, it's a surprising and undocumented behaviour. I might see if I can put together a simple repro and file a bug tomorrow.

For now, the simplest thing to do might be to simply not show the cumulative sums when in reordering mode - just continue to use `*ngFor="let objective of bucket!.objectives"` in the `*ngIf="showOrderButtons"` part of the template, don't specify the `cumulativeSum` or `bucketAllocationLimit` in the `app-objective` there, and make the objective component work correctly if these are not supplied (don't try to display the cumulative sum).

src/app/bucket/bucket.component.html
```diff
@@ -21,30 +21,34 @@
...
         [isReorderingEnabled]="showOrderButtons"
         [otherBuckets]="otherBuckets"
         (onMoveBucket)="moveObjective($event[0], $event[1], $event[2])"
         (onDelete)="deleteObjective($event)"
-        (onChanged)="onObjectiveChanged($event[0], $event[1])"></app-objective>
+        (onChanged)="onObjectiveChanged($event[0], $event[1])" ></app-objective>
     </mat-list-item>
 </mat-list>
 <mat-list dense *ngIf="showOrderButtons" cdkDropList (cdkDropListDropped)="reorderDrop($event)" class="app-mat-list">
-    <mat-list-item *ngFor="let objective of bucket!.objectives" cdkDrag>
```
amdw, line 37: As discussed elsewhere on this PR, this seems to break drag-and-drop.

For now, it might be better not to change this part of the template at all, and make the `app-objective` work correctly (don't show the cumulative sum) if `cumulativeSum` and `bucketAllocationLimit` are not supplied.

amdw (changes requested):
Many thanks! I think we're getting close to something we can merge.

src/app/bucket/bucket.component.ts
```diff
@@ -161,4 +160,26 @@ export class BucketComponent implements OnInit {
     let total = this.resourcesAllocated();
     return total ? this.committedResourcesAllocated() / total : 0;
   }
+
+displayObjectives():DisplayObjective[] {
```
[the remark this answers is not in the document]
r-sreesaran: I do use VS code. I will format things up using the formatter.

src/app/bucket/bucket.component.ts
```diff
@@ -161,4 +160,26 @@ export class BucketComponent implements OnInit {
...
+  
+  let displayObjectives:Array<DisplayObjective>=[];
+  let cumulativeSum =0;
+  for(let postion=0;postion< (this.bucket!.objectives)?.length;postion++ ) { 
+    let currentoObjective = this.bucket!.objectives[postion]
+     if (postion == 0) {
```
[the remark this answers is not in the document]
r-sreesaran: yes correct

src/app/bucket/bucket.component.ts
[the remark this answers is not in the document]
r-sreesaran: I will add a test.

r-sreesaran:
I have not added the test with the latest changes. I will write the test and push the changes.

r-sreesaran:
I was not able to execute the tests locally. 
"
 import firebase from 'firebase/app';
          ~~~~~~~~

  ../../../node_modules/firebase/index.d.ts:9546:1
    9546 export = firebase;
         ~~~~~~~~~~~~~~~~~~
    This module is declared with using 'export =', and can only be used with a default import when using the 'allowSyntheticDefaultImports' flag.
"

Any idea on how to resolve them ?. 
But I am able to execute tests from master branch. But my checkins dont have any related changes.

amdw:
For what it's worth, `npx ng test` succeeds for me when I run it against your latest commit de67590 (after merging the latest changes from the `master` branch).

amdw:
Also, just to note that I filed a bug against Angular Material for the problem with the drag-and-drop which we experienced here.

https://github.com/angular/components/issues/21682

I put together a minimal repro for the problem here:

https://stackblitz.com/github/amdw/drag-drop-repro

Please let me know when you'd like me to take another look at this PR. Thanks!

r-sreesaran:
after the merging the latest changes from master. The tests work I have added a single test as of now. 
I will add couple of other tests and let you know. 
Again many a thanks for all the review.

amdw:
For what it's worth, I found a workaround to the drag-and-drop problem. It works if you use a pipe in the `*ngFor` instead of a template function, like this:

https://github.com/amdw/drag-drop-repro/commit/fbe1161fb2313f9f04448d6e402f5c3dac1d6ddc

I think this would work well here, as the `bucket.objectives` array is immutable.

It's up to you whether you pursue this idea as part of this PR. It's not a big deal to me either way.

r-sreesaran:
I will make changes and push the code

r-sreesaran:
tests for newly created pipe is alone pending.

r-sreesaran:
Tests were added and also existing tests that were broken due introduction of new pipe was also fixed. 
@amdw  please do take a look and share your comments

amdw:
Thanks very much! I will merge this now. There are a few small tweaks I'd like to make, but it's probably quicker for me to do those myself now.

Issue #115, opened 2021-01-24
Use Prettier for code formatting


amdw:
I'd like to enforce code formatting rules using Prettier:

https://prettier.io/docs/en/install.html

amdw:
Currently enabled for everything except .html (for which I think the Prettier rules look a little weird).

I will consider removing this exclusion in the future. Closing the issue for now.

PR #121, opened 2021-01-27
Set up Prettier code formatting (#115)

amdw:


amdw:
Already enabled Prettier without HTML in another PR. I'll consider enabling HTML later.

Issue #31, opened 2020-06-10
Support rich formatting (e.g. markdown) in objectives and their notes


amdw:
There are occasions when it would be useful to have certain types of rich formatting, such as links or bold/italic text, in objectives and/or their notes.

This could be accomplished by supporting a markdown subset or some similar syntax. There is almost certainly already a suitable Javascript library for this.

amdw:
I should also note here that security would be a major consideration in deciding how to implement this.

Content would have to be put through some sort of rendering library in the client, and then the resulting HTML included in the DOM, through an `innerHTML` binding or similar. This raises the potential for XSS vulnerabilities.

I believe Angular has some sanitization functionality that would help mitigate against this. However, I'd want to seek a security review of the proposed approach before implementing if possible, just to make sure I haven't missed anything.

https://angular.io/guide/security#preventing-cross-site-scripting-xss

amdw:
I asked for some security advice internally regarding this feature.

I was told that we should be able to trust Angular's built-in sanitizer to mitigate XSS vulnerabilities that might otherwise arise from passing untrusted input to a Markdown library (most of which don't seem to sanitize their outputs -- e.g. [Marked docs](https://github.com/markedjs/marked#usage), [Showdown docs](https://github.com/showdownjs/showdown/wiki/Markdown's-XSS-Vulnerability-(and-how-to-mitigate-it))).

It was also suggested that I could consider coupling this with [DOMPurify in a restrictive configuration](https://github.com/cure53/DOMPurify/tree/main/demos#advanced-config-demo-link). This would add a further defence in depth against XSS vulnerabilities, as well as providing an easy way to restrict the Markdown features used, via a HTML tag allow-list.

The intent of this feature is basically just to enable the user to add very simple formatting like italics, bold and links; I wouldn't want to allow, for example, large heading tags to be injected in objectives by Markdown `##` directives or similar. So overall this seems like quite an attractive approach.

PR #148, opened 2021-04-16
Support basic Markdown in objectives (#31)

amdw:


amdw:
The one problem with this right now is that if you enable Markdown for an objective and add a link, the link will appear in the preview, but it won't be there when you OK out of the edit dialog, because the period will still be in edit mode.

Rather than eliminating links altogether in edit mode, it would be better to replace them with underlined text. That way, the user will see that their Markdown has not been ignored. If they try to click the link, it will reopen the edit dialog, which is the desired behaviour. Then, when the user disables edit mode, the link will be displayed and should work as expected.

amdw:
It would probably also be sensible to give the user some indication in the UI of what types of Markdown are supported.

PR #152, opened 2021-04-28
Update styling of Add/Edit Objective dialog

sttrinh:
- Increase form fields to 320px
- Move hints out of `<mat-label>` and into `hintLabel`
- Move help link out of `<mat-checkbox>` to fix focus
- Fix alignment and centering of checkbox, label, and icon

### DIFF
--- a/src/app/edit-objective-dialog/edit-objective-dialog.component.ts
+++ b/src/app/edit-objective-dialog/edit-objective-dialog.component.ts
@@ -12,7 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { Component, OnInit, Inject, EventEmitter } from '@angular/core';
+import {
+  Component,
+  EventEmitter,
+  Inject,
+  OnInit,
+  ViewEncapsulation,
+} from '@angular/core';
 import { MatDialogRef, MAT_DIALOG_DATA } from '@angular/material/dialog';
 import {
   CommitmentType,
@@ -114,6 +120,7 @@ const makeObjective = (edited: EditedObjective): ImmutableObjective =>
   selector: 'app-edit-objective-dialog',
   templateUrl: './edit-objective-dialog.component.html',
   styleUrls: ['./edit-objective-dialog.component.css'],
+  encapsulation: ViewEncapsulation.None,
 })
 export class EditObjectiveDialogComponent implements OnInit {
   constructor(
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -57,7 +57,6 @@ import { LoginComponent } from './login/login.component';
 import { NotificationService } from './services/notification.service';
 import { AuthInterceptor } from './services/auth.interceptor';
 import { firebaseConfig } from '../environments/firebaseConfig';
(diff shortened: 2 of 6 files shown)


src/app/edit-objective-dialog/edit-objective-dialog.component.ts
```diff
@@ -114,6 +120,7 @@ const makeObjective = (edited: EditedObjective): ImmutableObjective =>
   selector: 'app-edit-objective-dialog',
   templateUrl: './edit-objective-dialog.component.html',
   styleUrls: ['./edit-objective-dialog.component.css'],
+  encapsulation: ViewEncapsulation.None,
```
amdw, line 123: Could you please help me understand why this is necessary?

I tried removing this line and removing the `app-edit-objective-dialog` from the front of all the CSS clauses, and it seemed to work fine. Am I missing something?

amdw (commented):
Many thanks - this looks great! I especially like the hint labels and the flex layout fixes.

Just one question inline below.

src/app/edit-objective-dialog/edit-objective-dialog.component.ts
sttrinh: Angular performs view encapsulation by default, so for this component, Angular generates a unique attribute that is appended to the elements in the template, and also attaches that attribute to your CSS rules.

e.g. If you have the following:
```
<body>
  <h1>PeopleMath</h1>
</body>

body h1 {
  color: blue;
}
```

Then Angular will do something like this:
```
<body _ngcontent-sjh-c240>
  <h1 _ngcontent-sjh-c240>PeopleMath</h1>
</body>

div[_ngcontent-sjh-c240] h1[_ngcontent-sjh-c240] {
  color: blue;
}
```

This encapsulates all the rules in your CSS so that they only apply to your component and don't get mixed with other components.

The problem with this is you won't be able to override styles in child components **since Angular applies the attribute to the child selector as well**. e.g. I'm trying to target the `.mat-button-wrapper` class (child of `.mat-icon-button`), but since this element is nested within the `mat-icon-button`, Angular doesn't append this `_ngcontent-sjh-c240` attribute to children of `mat-icon-button`, and so the `.mat-button-wrapper[_ngcontent-sjh-c240]` selector won't match anything and the rule doesn't apply.

The following selectors won't work if view encapsulation is on:
- `.mat-checkbox-inner-container`
- `.mat-icon-button .mat-button-wrapper`

Having the `app-edit-objective-dialog` selector prefixed to every CSS rule is super redundant, but it's needed to preserve encapsulation when turning off Angular's encapsulation.

With SCSS you can simply nest them e.g.:
```
app-edit-objective-dialog {
  .people {}

  .math()
}
```

But since this is a regular CSS file, we have to do it the old-fashioned way for parent-child selectors.

amdw: Thank you very much for taking the time to write this extremely useful explanation!

amdw (approved):
Thanks very much for sending this PR! The objective editing dialog has a very nice layout now - I tested it on devices of several different sizes.

sttrinh:
No problem, glad it worked out!

PR #154, opened 2021-05-06
Update reordering CSS and expand clickable area for objectives

sttrinh:
Bucket: Update reordering style and animation
- Make list items more readable while dragging
- Show placeholder where list item will be dragged to
- Keep consistent styling between dragged and static items

Expand clickable area for objectives
- Refactor click logic to use event delegation, single listener for all clicks
- Reduce passing excessive state down from Bucket to Objective
- Add unsubscribe hooks for dialog close states
- Show grey background hover effect and pointer cursor to indicate clickable area
- Show underline on assignees' names to indicate clickability

### DIFF
--- a/src/app/bucket/bucket.component.css
+++ b/src/app/bucket/bucket.component.css
@@ -1,3 +1,7 @@
+app-bucket {
+  --color: rgba(0, 0, 0, 0.04);
+}
+
 app-bucket .header-button {
   float: right;
 }
@@ -10,7 +14,12 @@ app-bucket .mat-icon-button {
   width: 30px;
 }
 
-app-bucket .cdk-drop-list-dragging mat-list-item {
+app-bucket .interactive:hover {
+  background-color: var(--color);
+  cursor: pointer;
+}
+
+app-bucket .cdk-drop-list-dragging .interactive {
   background-color: #fff;
   cursor: move;
 }
@@ -20,7 +29,7 @@ app-bucket .mat-list-item-content {
 }
 
 app-bucket .cdk-drag-placeholder {
-  background-color: rgba(0, 0, 0, 0.04);
+  background-color: var(--color);
   cursor: move;
   min-height: 40px;
   transition: var(--transition); /* From <body> */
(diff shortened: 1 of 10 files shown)


amdw (changes requested):
Many thanks! I merged the first commit, as that seems fairly straightforward, and is a clear win for the drag-and-drop reordering.

The second commit looks quite a bit more involved!

I do like the new mouseover / click aesthetics, though I see a few problems:

* Most importantly, it looks like the "move bucket" functionality is now broken. If you edit an objective, and then click the "move" button to move it to a different bucket, it no longer works: nothing happens, and you get a "Cannot read property 'emit' of undefined" error in the console. I imagine it has something to do with the state refactoring.
* We're now showing hover-underline on assignees even when we're not in edit mode, where clicking them won't do anything.
* There's a space before the first assignee that's now getting underlined on mouseover, which looks slightly ugly.

Those things are probably fixable (we've managed to avoid the latter two pitfalls with the "Assign" button you get when there are no assignments), but even then, my initial reaction is that we're paying quite a big complexity price for this aesthetic improvement.

The "move bucket" bug is an example of the kind of breakage that can result from that kind of complexity.

Perhaps I'm misjudging, and the problem is mainly a lack of test coverage... interested in your thoughts. Thanks!

sttrinh:
Those 3 points you mentioned should all be pretty straightforward to fix. Since we are about 80% of the way there in terms of feature complete for this PR, it wouldn't take much more to fix the bugs you identified and add more tests to cover all our bases.

I agree with your assessment of the complexity though. As I was making the changes I realized I was making way more changes than I anticipated, and was not sure if I should keep going.

As you are the primary maintainer of the code, I think it all boils down to whether you would feel comfortable supporting this different way of event binding. If not, it's perfectly understandable and we don't have to merge this in order to avoid any potential consequences and we maintain consistency across the app. So I'm slightly leaning towards not merging this since it seems like a big tradeoff.

amdw:
Many thanks! If the only benefit we get from this is a better clickable area on objectives and the grey highlighting, my sense is that the asking price is a bit too high and we should probably stop here. Is that OK with you?

Again, really appreciate the time you took to send this PR - the first commit is merged already.

sttrinh:
Yup totally ok, thanks! I'll send more PRs later if I think of something

Issue #132, opened 2021-02-05
Centre loading spinners


amdw:
A minor aesthetic issue: it would be nice if the loading spinners were centred on the screen, rather than jammed into the top left with no margin.

rojanDinc:
Hi I would like to get assigned to this.

amdw:
Many thanks - please go ahead!

PR #169, opened 2021-06-22
Add progress spinner component

rojanDinc:
Closing #132


rojanDinc:
> Thanks for your pull request. It looks like this may be your first contribution to a Google open source project (if not, look below for help). Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> 📝 **Please visit https://cla.developers.google.com/ to sign.**
> 
> Once you've signed (or fixed any issues), please reply here with `@googlebot I signed it!` and we'll verify it.
> 
> #### What to do if you already signed the CLA
> ##### Individual signers
> * It's possible we don't have your GitHub username or you're using a different email address on your commit. Check [your existing CLA data](https://cla.developers.google.com/clas) and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> 
> ##### Corporate signers
> * Your company has a Point of Contact who decides which employees are authorized to participate. Ask your POC to be added to the group of authorized contributors. If you don't know who your Point of Contact is, direct the Google project maintainer to [go/cla#troubleshoot](http://go/cla#troubleshoot) ([Public version](https://opensource.google/docs/cla/#troubleshoot)).
> * The email used to register you as an authorized contributor must be the email used for the Git commit. Check [your existing CLA data](https://cla.developers.google.com/clas) and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> * The email used to register you as an authorized contributor must also be [attached to your GitHub account](https://github.com/settings/emails).
> 
> ℹ️ **Googlers: [Go here](https://goto.google.com/prinfo/https%3A%2F%2Fgithub.com%2Fgoogle%2Fpeoplemath%2Fpull%2F169) for more info**.

@googlebot I signed it!

rojanDinc:
> We found a Contributor License Agreement for you (the sender of this pull request), but were unable to find agreements for all the commit author(s) or Co-authors. If you authored these, maybe you used a different email address in the git commits than was used to sign the CLA ([login here](https://cla.developers.google.com/) to double check)? If these were authored by someone else, then they will need to sign a CLA as well, and confirm that they're okay with these being contributed to Google.
> In order to pass this check, please resolve this problem and then comment `@googlebot I fixed it.`. If the bot doesn't comment, it means it doesn't think anything has changed.
> 
> ℹ️ **Googlers: [Go here](https://goto.google.com/prinfo/https%3A%2F%2Fgithub.com%2Fgoogle%2Fpeoplemath%2Fpull%2F169) for more info**.

@googlebot I fixed it.

amdw (changes requested):
Many thanks for the pull request.

Could you please explain the benefit of introducing a new component for this?

There is already a `mat-spinner` component provided by Angular Material, so as far as I can see we should not need to introduce any new components here.

We should be able to achieve the desired positioning just by applying CSS styles to the `mat-spinner`s that we already have. I would expect the whole issue should be possible to close just by modifying existing CSS files.

Could you please let me know if you see a problem with that approach?

rojanDinc:
> Many thanks for the pull request.
> 
> Could you please explain the benefit of introducing a new component for this?
> 
> There is already a `mat-spinner` component provided by Angular Material, so as far as I can see we should not need to introduce any new components here.
> 
> We should be able to achieve the desired positioning just by applying CSS styles to the `mat-spinner`s that we already have. I would expect the whole issue should be possible to close just by modifying existing CSS files.
> 
> Could you please let me know if you see a problem with that approach?

I was considering the reusability of the component and the ease of changing the underlying spinner if that would be needed in the future. But you're right I could introduce a global CSS class which all `mat-spinner` could use if they should be centered but they would need to be wrapped by a `div` to get the desired outcome. I could go ahead and implement your suggestion. What are your thoughts?

amdw:
Thanks for explaining! I think we're OK not to have a separate component for now - we can just stick with `mat-spinner`. I think it's very unlikely we'd move away from `mat-spinner` in the future, and if we do, we can always create a new component at that point.

I think it should be possible to make the necessary styling changes purely by putting a `mat-spinner { ... }` clause into the CSS files of the components that use the spinner. If this isn't possible, then adding a `div` around the spinner in order to apply additional styles is also fine.

rojanDinc:
Yes I managed to center the `mat-spinner` by using a css class I created called `centered`. This css class should work for any element that should be centered. Im against forcing all `mat-spinner`s to be centered I think it should be decided with a css class, what do you think?

src/styles.css
```diff
@@ -41,3 +41,10 @@ body {
 .cdk-drop-list-dragging .cdk-drag {
   transition: var(--transition);
 }
+
+.centered {
+  position: fixed;
```
amdw, line 46: This appears to be overridden by a `position: relative` on the `mat-spinner` itself. I don't think this is important, and it can be removed. I'll follow up and do this after merging.

amdw (approved):
This looks good, thanks! One minor comment which I can take care of after merging.

Issue #258, opened 2023-11-24
Add accessibility template ESLint rules


amdw:
Add the following to .eslintrc.json for `*.html`:

`plugin:@angular-eslint/template/accessibility`

PR #290, opened 2024-09-27
Add accessibility linting (#258)

Nolle10:


amdw:
Unfortunately this change breaks the build. We can't add a new lint check until all errors are resolved.

Apologies for the delay in reviewing this PR.

amdw:
Closing this PR for now as it would break the build.

Please feel free to resubmit this change if we can get it to a stage where there are no lint errors.

Issue #291, opened 2024-09-28
Add tooltips for better usability


Nolle10:
Add tooltips to editing blocks and reordering buttons, for better accessibility and usability.

PR #292, opened 2024-09-28
Add aria labels (#291)

Nolle10:
This PR enhances the usability of the user interface by adding tooltips to the action buttons in the editing section. The tooltips provide clear, concise descriptions of each button's function, reducing confusion and improving the overall user experience.

**Key Changes:**
Added `matTooltip `attributes to the following buttons:
- Toggle Block Editing: Indicates that this button enables or disables block editing mode.
- Toggle Reordering: Clarifies that this button allows the user to reorder items in the list.

**Benefits:**
Improves user understanding of button functionalities.
Enhances accessibility for users who may not be familiar with the UI.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,11 +1,18 @@
 {
   "root": true,
-  "ignorePatterns": ["projects/**/*"],
+  "ignorePatterns": [
+    "projects/**/*"
+  ],
   "overrides": [
     {
-      "files": ["*.ts"],
+      "files": [
+        "*.ts"
+      ],
       "parserOptions": {
-        "project": ["tsconfig.json", "e2e/tsconfig.e2e.json"],
+        "project": [
+          "tsconfig.json",
+          "e2e/tsconfig.e2e.json"
+        ],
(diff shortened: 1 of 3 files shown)


amdw:
Apologies for the delay in reviewing this pull request.

I tried this out and I saw some behaviour I can't explain.

The first time you click the "edit" button and hover over one of the changed buttons, the tooltip appears exactly as you'd expect, right above the button:

good

However if you then click the tick (to exit edit mode), so the buttons disappear, and then click "edit" again and hover over the buttons, then the tooltips appear in the very top left of the window, which looks really strange and confusing:

bad

This happens consistently after edit mode has been entered and exited more than once.

Do you have any idea why this happens?

I don't think I can merge a PR which has this effect. Do you know how to fix it so the tooltip consistently appears in the right place?

If not, then if the only consideration is accessibility, might it be sufficient to simply add an `aria-description` to the buttons instead of a tooltip?

amdw:
Closing for now - please feel free to reopen if the problem mentioned here can be addressed. Thanks!

